PR #2, opened 2015-08-18
(TK-258) Config checking and transformation functions

jeffmccune:
NOTE:

The ticket describes implementing the protocol function to wrap the middleware.
This work remains for this ticket and I'm in the process of implementing that
now.  I'm getting the PR up for early review, however, in an effort to get this
in before the sprint ends.

---

This patch establishes the basic structure of the TK service and implements
initialization logic that does the following:
- pull in a "rules" vector from the "authorization" Trapperkeeper
  configuration section (:ConfigService get-config)
- do any initial structural validation that seems appropriate
- transform that structure into the working
  puppetlabs.trapperkeeper.authorization.rules.Rules vector that
  "trapperkeeper-authorization" will need to use for its allowed? checks
  later on
- store the Rules vector into the service context

(TK-258) Add transform-config-rule-to-acl function

Without this patch there is no way to transform a rule expressed as
trapperkeeper configuration into an Access Control List data structure.
This is a problem because we need to feed ACL objects into the
authorization library.  This patch addresses the problem by implementing
the transformation function.  The implementation starts with an empty acl,
which itself is a sorted set, and then builds up the ACL by adding named
rules for each allow and deny entry in the TK configuration.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -33,11 +33,9 @@
   ;; code that we have.
   :classifiers [["test" :testutils]]
 
-  :profiles {:dev {:dependencies [[spyscope "0.1.4"]
-                                  [puppetlabs/trapperkeeper ~tk-version :classifier "test" :scope "test"]
+  :profiles {:dev {:dependencies [[puppetlabs/trapperkeeper ~tk-version :classifier "test" :scope "test"]
                                   [puppetlabs/trapperkeeper-webserver-jetty9 ~tk-jetty-version :classifier "test"]
-                                  [puppetlabs/kitchensink "1.0.0" :classifier "test"]]
-                   :injections [(require 'spyscope.core)]}
+                                  [puppetlabs/kitchensink "1.0.0" :classifier "test"]]}
              :testutils {:source-paths ^:replace ["test"]}}
 
   ;; this plugin is used by jenkins jobs to interrogate the project version
--- a/project.clj
+++ b/project.clj
@@ -1,3 +1,5 @@
+(def tk-version "1.1.1")
+(def tk-jetty-version "1.3.1")
 (def tk-authz-version "0.0.1")
 
 (defproject masterzen/trapperkeeper-authorization tk-authz-version
@@ -12,15 +14,17 @@
   :pedantic? :abort
 
   :dependencies [[org.clojure/clojure "1.6.0"]
+                 [puppetlabs/trapperkeeper ~tk-version]
+                 [puppetlabs/trapperkeeper-webserver-jetty9 ~tk-jetty-version]
                  ;; Logging
                  [org.clojure/tools.logging "0.2.6"]
                  ;; Filesystem utilities
                  [me.raynes/fs "1.4.5"]
                  [org.clojure/tools.cli "0.3.0"]
-                 [prismatic/schema "0.2.2"]
+                 [prismatic/schema "0.4.0"]
                  [inet.data "0.5.5"]
                  [clj-time "0.5.1"]
-                 [puppetlabs/typesafe-config "0.1.1"]
+                 [puppetlabs/typesafe-config "0.1.4"]
                  [puppetlabs/ssl-utils "0.8.0"]]
 
   ;; By declaring a classifier here and a corresponding profile below we'll get an additional jar
@@ -30,6 +34,8 @@
   :classifiers [["test" :testutils]]
 
   :profiles {:dev {:dependencies [[spyscope "0.1.4"]
+                                  [puppetlabs/trapperkeeper ~tk-version :classifier "test" :scope "test"]
+                                  [puppetlabs/trapperkeeper-webserver-jetty9 ~tk-jetty-version :classifier "test"]
                                   [puppetlabs/kitchensink ~"1.0.0" :classifier "test"]]
                    :injections [(require 'spyscope.core)]}
              :testutils {:source-paths ^:replace ["test"]}}
--- a/project.clj
+++ b/project.clj
@@ -2,9 +2,9 @@
 (def tk-jetty-version "1.3.1")
 (def tk-authz-version "0.0.1")
 
-(defproject masterzen/trapperkeeper-authorization tk-authz-version
+(defproject puppetlabs/trapperkeeper-authorization tk-authz-version
   :description "TrapperKeeper authorization system"
-  :url "http://github.com/masterzen/trapperkeeper-authorization"
+  :url "http://github.com/puppetlabs/trapperkeeper-authorization"
   :license {:name "Apache License, Version 2.0"
             :url "http://www.apache.org/licenses/LICENSE-2.0.html"}
 
--- a/project.clj
+++ b/project.clj
@@ -36,7 +36,7 @@
   :profiles {:dev {:dependencies [[spyscope "0.1.4"]
                                   [puppetlabs/trapperkeeper ~tk-version :classifier "test" :scope "test"]
                                   [puppetlabs/trapperkeeper-webserver-jetty9 ~tk-jetty-version :classifier "test"]
-                                  [puppetlabs/kitchensink ~"1.0.0" :classifier "test"]]
+                                  [puppetlabs/kitchensink "1.0.0" :classifier "test"]]
                    :injections [(require 'spyscope.core)]}
              :testutils {:source-paths ^:replace ["test"]}}
 
--- a/project.clj
+++ b/project.clj
@@ -23,9 +23,9 @@
                  [org.clojure/tools.cli "0.3.0"]
                  [prismatic/schema "0.4.0"]
                  [inet.data "0.5.5"]
-                 [clj-time "0.5.1"]
+                 [clj-time "0.7.0"]
                  [puppetlabs/typesafe-config "0.1.4"]
-                 [puppetlabs/ssl-utils "0.8.0"]]
+                 [puppetlabs/ssl-utils "0.8.1"]]
 
   ;; By declaring a classifier here and a corresponding profile below we'll get an additional jar
   ;; during `lein jar` that has all the code in the test/ directory. Downstream projects can then
--- a/src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
@@ -15,7 +15,7 @@
 
 (def required-or-key
   "At least one of these keys is required in an auth rule map."
-  [:deny :allow])
+  #{:deny :allow})
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Private
--- a/src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
@@ -1,0 +1,126 @@
+(ns puppetlabs.trapperkeeper.services.authorization.authorization-core
+  (:require [clojure.string :as str]
+            [puppetlabs.kitchensink.core :as ks]
+            [schema.core :as schema]
+            [puppetlabs.trapperkeeper.authorization.rules :as rules]
+            [puppetlabs.trapperkeeper.authorization.acl :as acl])
+  (:import (java.util.regex PatternSyntaxException)))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Constants
+
+(def required-keys
+  "Keys required in an auth rule map."
+  [:path :type])
+
+(def required-or-key
+  "At least one of these keys is required in an auth rule map."
+  [:deny :allow])
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Private
+
+(defn pprint-rule
+  [rule]
+  (str/trim (ks/pprint-to-string rule)))
+
+
+(schema/defn transform-config-rule-to-acl :- acl/ACL
+  [{:keys [allow deny] :as config-rule}]
+  (let [type (if (= "regex" (.toLowerCase (:type config-rule)))
+               :regex :domain)
+        base {:qualifier :exact
+              :length nil
+              :type type}
+        allow-base (assoc base :auth-type :allow)
+        deny-base (assoc base :auth-type :deny)
+        allow (if (string? allow) [allow] allow)
+        deny (if (string? deny) [deny] deny)]
+    (set (concat
+           (map #(assoc allow-base :pattern %) allow)
+           (map #(assoc deny-base :pattern %) deny)))))
+
+(schema/defn transform-config-rule :- rules/Rule
+  [config-rule]
+  (let [type (name (.toLowerCase (:type config-rule)))]
+    {:type (if (= type "regex") :regex :string)
+     :path (re-pattern (:path config-rule))
+     :method :any
+     :acl (transform-config-rule-to-acl config-rule)}))
+
+(defn validate-auth-config-rule!
+  "Tests to see if the given map contains the proper data to define an auth
+  rule. Returns the provided rule if successful, otherwise throws an exception
+  with a useful error message."
+  [rule]
+  (when-not (map? rule)
+    (throw (IllegalArgumentException. "An authorization rule should be specified as a map")))
+  (let [rule-keys (keys rule)]
+    (doseq [k required-keys]
+      (when-not (some #(= k %) rule-keys)
+        (throw (IllegalArgumentException. (str "The authorization rule specified as "
+                                               (pprint-rule rule)
+                                               " does not contain a '" (name k) "' key.")))))
+    (when-not (some #(.contains required-or-key %) rule-keys)
+      (throw (IllegalArgumentException.
+               (str "Authorization rule specified as  "
+                    (pprint-rule rule)
+                    " must contain either a 'deny' or 'allow' rule.")))))
+  (when-not (string? (:type rule))
+    (throw (IllegalArgumentException.
+             (str "The type set in the authorization rule specified "
+                  "as " (pprint-rule rule) " should be a "
+                  "string that is either 'path' or 'regex'."))))
+  (let [type (.toLowerCase (:type rule))]
+    (when-not (or (= type "path") (= type "regex"))
+      (throw (IllegalArgumentException.
+               (str "The type set in the authorization rule specified "
+                    "as " (pprint-rule rule) " is invalid. "
+                    "It should be set to either 'path' or 'regex'.")))))
+  (let [{path :path} rule]
+    (when-not (string? path)
+      (throw (IllegalArgumentException.
+               (str "The path set in the authorization rule specified as "
+                    (pprint-rule rule) " is invalid. It should be "
+                    "a string.")))))
+  (when (= (name (:type rule)) "regex")
+    (try
+      (re-pattern (:path rule))
+      (catch PatternSyntaxException e
+        (throw (IllegalArgumentException.
+                 (str "The path regex provided in the rule defined as "
+                      (pprint-rule rule) " is invalid: "
+                      (.getMessage e)))))))
+  (doseq [[type names] (select-keys rule [:allow :deny])]
+    (if (vector? names)
+      (when-not (every? string? names)
+        (throw (IllegalArgumentException.
+                 (str "The " (name type) " list in the rule specified as "
+                      (pprint-rule rule)
+                      " contains one or more names that are not strings."))))
+      (when-not (string? names)
+        (throw (IllegalArgumentException.
+                 (str "The name '" names "' in the '" (name type) "' field of the "
+                      "rule specified as " (pprint-rule rule) " is not a string."))))))
+  rule)
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Public
+
+(defn validate-auth-config!
+  "Validates the given authorization service configuration. If an invalid
+  configuration is found an IllegalArgumentException will be thrown, otherwise
+  the input config will be returned."
+  [config]
+  (when-not (vector? config)
+    (throw (IllegalArgumentException.
+             "The providede authorization service config is not a list.")))
+  (doseq [rule config]
+    (validate-auth-config-rule! rule))
+  config)
+
+(schema/defn transform-config :- rules/Rules
+  "Transforms the authorization service config into a list of Rules that work
+  with the authorization code."
+  [config]
+  (map transform-config-rule config))
--- a/src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
@@ -24,7 +24,6 @@
   [rule]
   (str/trim (ks/pprint-to-string rule)))
 
-
 (schema/defn transform-config-rule-to-acl :- acl/ACL
   [{:keys [allow deny] :as config-rule}]
   (let [type (if (= "regex" (.toLowerCase (:type config-rule)))
--- a/src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
@@ -25,19 +25,12 @@
   (str/trim (ks/pprint-to-string rule)))
 
 (schema/defn transform-config-rule-to-acl :- acl/ACL
-  [{:keys [allow deny] :as config-rule}]
-  (let [type (if (= "regex" (.toLowerCase (:type config-rule)))
-               :regex :domain)
-        base {:qualifier :exact
-              :length nil
-              :type type}
-        allow-base (assoc base :auth-type :allow)
-        deny-base (assoc base :auth-type :deny)
-        allow (if (string? allow) [allow] allow)
+  [{:keys [allow deny]}]
+  (let [allow (if (string? allow) [allow] allow)
         deny (if (string? deny) [deny] deny)]
-    (set (concat
-           (map #(assoc allow-base :pattern %) allow)
-           (map #(assoc deny-base :pattern %) deny)))))
+    (let [allow-acl (reduce #(acl/allow %1 %2) acl/empty-acl allow)
+          full-acl (reduce #(acl/deny %1 %2) allow-acl deny)]
+      full-acl)))
 
 (schema/defn transform-config-rule :- rules/Rule
   [config-rule]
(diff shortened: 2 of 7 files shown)


src/puppetlabs/trapperkeeper/services/authorization/authorization_service.clj
```diff
@@ -0,0 +1,23 @@
...
+  [[:ConfigService get-in-config]]
+  (init
+    [this context]
+    (let [config (get-in-config [:authorization :rules])]
+      (validate-auth-config! config)
+      (assoc context :authorization {:rules (transform-config config)})))
+
+  (wrap-handler
+    [this handler]
+    (let [context (service-context this)
```
jeffmccune: @camlow325 @nwolfe Would appreciate your help with this one...  I'm at the point where I'm trying to finish the last item of TK-258, which is to actually implement the TK service function that accepts a ring request handler and returns a new handler wrapped with the authorization check.

I'm trying to obtain the service context using `service-context` as mentioned at https://github.com/puppetlabs/trapperkeeper/wiki/Defining-Services  but I keep running into the error:

```
CompilerException java.lang.RuntimeException: Unable to resolve symbol: service-context in this context, compiling:(src/puppetlabs/trapperkeeper/services/authorization/authorization_service.clj:21:19) 
```

Any ideas?  Looking at other uses, it seems to be correct so I think I'm definitely overlooking something.

jeffmccune: Nevermind, a community member filed an issue and there's the answer at https://github.com/puppetlabs/trapperkeeper/issues/166

src/puppetlabs/trapperkeeper/services/authorization/authorization_service.clj
```diff
@@ -0,0 +1,24 @@
+(ns puppetlabs.trapperkeeper.services.authorization.authorization-service
+  (:require [puppetlabs.trapperkeeper.core :refer [defservice]]
+            [puppetlabs.trapperkeeper.services :refer [service-context]]
+            [puppetlabs.trapperkeeper.authorization.ring-middleware :refer
+             [wrap-authorization-check]]
+            [puppetlabs.trapperkeeper.services.authorization.authorization-core :refer :all]))
+
+(defprotocol AuthorizationService
+  (wrap-handler [this handler]))
```
jeffmccune: What do you think of renaming `wrap-handler` to `wrap-with-authorization-check`?  Looking at the integration test, `(wrap-handler handler)` seems a bit ambiguous.

project.clj
```diff
@@ -1,3 +1,5 @@
+(def tk-version "1.1.1")
+(def tk-jetty-version "1.3.1")
 (def tk-authz-version "0.0.1")
 
 (defproject masterzen/trapperkeeper-authorization tk-authz-version
```
camlow325: We should change this to "puppetlabs/trapperkeeper-authorization" now.

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -0,0 +1,118 @@
...
+;;; Public
+
+(defn validate-auth-config!
+  "Validates the given authorization service configuration. If an invalid
+  configuration is found an IllegalArgumentException will be thrown, otherwise
+  the input config will be returned."
+  [config]
+  (when-not (vector? config)
+    (throw (IllegalArgumentException.
+             "The providede authorization service config is not a list.")))
```
camlow325: s/providede/provided

project.clj
jeffmccune: Fixed up in 6c5882f79c022edb2416537b213bc9b8c61f8b48

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
jeffmccune: Fixed up in 6c5882f79c022edb2416537b213bc9b8c61f8b48

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -0,0 +1,118 @@
...
+
+(schema/defn transform-config-rule :- rules/Rule
+  [config-rule]
+  (let [type (name (.toLowerCase (:type config-rule)))]
+    {:type (if (= type "regex") :regex :string)
+     :path (re-pattern (:path config-rule))
+     :method :any
+     :acl (transform-config-rule-to-acl config-rule)}))
+
+(defn validate-auth-config-rule!
```
camlow325, line 43: Just as a general comment, I think it would be cleaner to maintain this code if we were using Prismatic Schema to do the heavy lifting of the validations rather than all of the hand-rolled stuff that is here.  That said, though, I understand how onerous it is for a user to interpret the error messages from a Schema validation failure and that the code here may be an attempt to compensate for those and create much more readable failures for users.  I wouldn't block the PR based on this, but it's probably worth revisiting at some point later on.  As it is, I expect that we'll have a fair bit of turmoil here in follow-on PRs since the format of the rule definitions are still a little in flux.

src/puppetlabs/trapperkeeper/services/authorization/authorization_service.clj
camlow325: `wrap-with-authorization-check` sounds like a better name to me as well.  I'd be good with that change.

src/puppetlabs/trapperkeeper/services/authorization/authorization_service.clj
```diff
@@ -0,0 +1,24 @@
...
+
+(defservice authorization-service
+  AuthorizationService
+  [[:ConfigService get-in-config]]
+  (init
+    [this context]
+    (let [config (get-in-config [:authorization :rules])]
+      (validate-auth-config! config)
+      (assoc-in context [:authorization :rules] (transform-config config))))
```
camlow325: Not sure we need to nest the transformed config under ":authorization" since the context is specific to this service anyway, which is all about authorization.  Could just put it under ":rules" at the top-level.

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -0,0 +1,118 @@
...
+      (throw (IllegalArgumentException.
+               (str "Authorization rule specified as  "
+                    (pprint-rule rule)
+                    " must contain either a 'deny' or 'allow' rule.")))))
+  (when-not (string? (:type rule))
+    (throw (IllegalArgumentException.
+             (str "The type set in the authorization rule specified "
+                  "as " (pprint-rule rule) " should be a "
+                  "string that is either 'path' or 'regex'."))))
+  (let [type (.toLowerCase (:type rule))]
```
camlow325: Do we need to convert this to lower case here?  Seems like we could just choose to recognize only the lower-case form as valid and then not have to add code elsewhere that converts it to lowercase?  Not sure we've been supporting mixed case elsewhere in our TK services?

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -0,0 +1,118 @@
...
+  [{:keys [allow deny]}]
+  (let [allow (if (string? allow) [allow] allow)
+        deny (if (string? deny) [deny] deny)]
+    (let [allow-acl (reduce #(acl/allow %1 %2) acl/empty-acl allow)
+          full-acl (reduce #(acl/deny %1 %2) allow-acl deny)]
+      full-acl)))
+
+(schema/defn transform-config-rule :- rules/Rule
+  [config-rule]
+  (let [type (name (.toLowerCase (:type config-rule)))]
```
camlow325: If we decide to treat non-lower case as "invalid" in `validate-auth-config-rule!`, could avoid the `.toLowerCase` call here.

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -0,0 +1,118 @@
...
+    {:type (if (= type "regex") :regex :string)
+     :path (re-pattern (:path config-rule))
+     :method :any
```
camlow325, line 40: Do we not have support yet for mapping the `:method` from the config file to a `method` in the internal rule?

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
nwolfe: Also we should prefer the `clojure.string` functions like `lower-case` instead of calling methods on `String` directly.

I know we haven't been consistent with this in the past though.

project.clj
```diff
@@ -30,6 +34,8 @@
   :classifiers [["test" :testutils]]
 
   :profiles {:dev {:dependencies [[spyscope "0.1.4"]
+                                  [puppetlabs/trapperkeeper ~tk-version :classifier "test" :scope "test"]
+                                  [puppetlabs/trapperkeeper-webserver-jetty9 ~tk-jetty-version :classifier "test"]
                                   [puppetlabs/kitchensink ~"1.0.0" :classifier "test"]]
```
nwolfe: Whoa, what is this version? Don't think I've seen this before...is this a typo?

project.clj
```diff
@@ -12,15 +14,17 @@
...
                  ;; Filesystem utilities
                  [me.raynes/fs "1.4.5"]
                  [org.clojure/tools.cli "0.3.0"]
-                 [prismatic/schema "0.2.2"]
+                 [prismatic/schema "0.4.0"]
                  [inet.data "0.5.5"]
                  [clj-time "0.5.1"]
-                 [puppetlabs/typesafe-config "0.1.1"]
+                 [puppetlabs/typesafe-config "0.1.4"]
                  [puppetlabs/ssl-utils "0.8.0"]]
```
nwolfe: FYI we have `0.8.1` out, though it bumps the `clj-time` dependency to `0.7.0` and none of the other TK libs are on that version so we end up having to do some exclusions.

project.clj
```diff
@@ -30,6 +34,8 @@
   :classifiers [["test" :testutils]]
 
   :profiles {:dev {:dependencies [[spyscope "0.1.4"]
```
nwolfe: FYI we don't need to put `spyscope` in every single project anymore, you can leave this in your personal `profiles.clj`.

Doesn't hurt being here, but we're just dragging this stuff around and I've not heard of anyone using this other than Chris.

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -0,0 +1,118 @@
...
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Constants
+
+(def required-keys
+  "Keys required in an auth rule map."
+  [:path :type])
+
+(def required-or-key
+  "At least one of these keys is required in an auth rule map."
+  [:deny :allow])
```
nwolfe: We had code like this in puppet-server and for some reason the datatype was a vector instead of a set, even though we thought of it and treated like one.  We eventually changed them to sets, might do the same thing here.

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -0,0 +1,118 @@
...
+(defn pprint-rule
+  [rule]
+  (str/trim (ks/pprint-to-string rule)))
+
+(schema/defn transform-config-rule-to-acl :- acl/ACL
+  [{:keys [allow deny]}]
+  (let [allow (if (string? allow) [allow] allow)
+        deny (if (string? deny) [deny] deny)]
+    (let [allow-acl (reduce #(acl/allow %1 %2) acl/empty-acl allow)
+          full-acl (reduce #(acl/deny %1 %2) allow-acl deny)]
```
nwolfe: The anonymous functions are unnecessary, we can just do `(reduce acl/allow ...)` and `(reduce acl/deny ...)`

src/puppetlabs/trapperkeeper/services/authorization/authorization_service.clj
nwolfe: > since the context is specific to this service 

What does the function `(service-context this)` do then?

nwolfe: > What does the function (service-context this) do then?

Ohhhh right it allows you to get the service-specific context _from outside the service definition_.  Service lifecycle functions are already given the service-specific context...right.

test/puppetlabs/trapperkeeper/services/authorization/authorization_core_test.clj
```diff
@@ -0,0 +1,138 @@
+(ns puppetlabs.trapperkeeper.services.authorization.authorization-core-test
+  (:require [clojure.test :refer :all]
+            [puppetlabs.trapperkeeper.services.authorization.authorization-core :refer :all]
+            [schema.test]))
```
nwolfe, line 4: Huh, cool, guess we don't have to always tack on the `:as schema.test` part.  So used to just always doing a few keywords at the end :smile:

src/puppetlabs/trapperkeeper/services/authorization/authorization_service.clj
nwolfe: Yeah, something more than just `wrap-handler` would be good.

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
camlow325: It looks like the original intent in the tk-auth library was to have `path` be assumed as a default when `type` wasn't specified, essentially making `type` optional - https://github.com/puppetlabs/trapperkeeper-authorization/blob/1ac74458d6ee857f0fdcbf198584f635ed0ed1bd/src/puppetlabs/trapperkeeper/authorization/file/config.clj#L30.  This code would make `path` required in the rule.  In Puppet's auth.conf, the "path" expression itself controlled whether a literal or regex was used, so there wasn't really a notion of a default like there is where `type` is a separate element.  I'm okay with it being required as the code has here - makes things more explicit - and it isn't obvious to me whether literal or regex expressions would be used more often.  Figured I'd point that out in case others had thoughts on that.

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
nwolfe, line 43: I don't think it matters much right now, but we started down this style of validation with the CA service where the `validate!` function would throw all sorts of different exceptions.

We ended up wanting to catch those exceptions in the HTTP layer, and at that point we regretted having the validate function throw exceptions since we then had to have a bunch of catch statements for the different types.  

In the end it wasn't a big deal, but I think we kind of wish we had written validate to instead return a collection of errors rather than assuming we wanted to always throw them.

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -0,0 +1,118 @@
...
+      full-acl)))
+
+(schema/defn transform-config-rule :- rules/Rule
```
camlow325, line 35: It feels to me like this function is reimplementing a lot of the functionality that already exists in tk-authorization for creating a rule.  What if instead we were to create a function similar to `config->rule` (https://github.com/puppetlabs/trapperkeeper-authorization/blob/1ac74458d6ee857f0fdcbf198584f635ed0ed1bd/src/puppetlabs/trapperkeeper/authorization/file/config.clj#L66-L74) which just starts from a config as it would be pulled from the TK service instead of a HOCON config object - reusing backing functions like `build-rule` and `add-acl`.  I'm thinking if we did that, we could get rid of the logic that builds up the rule map, including the entire `transform-config-rule-to-acl` function.  Does that make sense?

src/puppetlabs/trapperkeeper/services/authorization/authorization_service.clj
```diff
@@ -0,0 +1,24 @@
...
+
+  (wrap-handler
+    [this handler]
+    (if-let [rules (get-in (service-context this) [:authorization :rules])]
+      (wrap-authorization-check handler rules)
+      (throw (IllegalStateException. (str "ERROR: No rules loaded."))))))
```
camlow325: Do we need to throw this exception here?  Didn't think we'd have a case where a user of the service protocol would be able to get to this since the `init` function would always have put the rules into the `context` before this function could be called.

src/puppetlabs/trapperkeeper/services/authorization/authorization_service.clj
jeffmccune: Yep

src/puppetlabs/trapperkeeper/services/authorization/authorization_service.clj
jeffmccune: I put it there because I mistakenly thought `(transform-config nil)` could return `nil`, but it doesn't:

```
(transform-config nil)
```

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -0,0 +1,118 @@
...
+  [config-rule]
+  (let [type (name (.toLowerCase (:type config-rule)))]
+    {:type (if (= type "regex") :regex :string)
+     :path (re-pattern (:path config-rule))
+     :method :any
+     :acl (transform-config-rule-to-acl config-rule)}))
+
+(defn validate-auth-config-rule!
+  "Tests to see if the given map contains the proper data to define an auth
+  rule. Returns the provided rule if successful, otherwise throws an exception
```
camlow325, line 45: I'd be fine with this just returning `nil` or "indeterminate" for the case that the validation is successful but round-tripping the rule seems okay as well.  Just looked like we ignored the return value in the production code for this anyway.

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -0,0 +1,118 @@
...
+        deny (if (string? deny) [deny] deny)]
+    (let [allow-acl (reduce #(acl/allow %1 %2) acl/empty-acl allow)
+          full-acl (reduce #(acl/deny %1 %2) allow-acl deny)]
+      full-acl)))
+
+(schema/defn transform-config-rule :- rules/Rule
+  [config-rule]
+  (let [type (name (.toLowerCase (:type config-rule)))]
+    {:type (if (= type "regex") :regex :string)
+     :path (re-pattern (:path config-rule))
```
camlow325, line 39: Another advantage to re-using parts of `build-rule` to do this is the handling for setting `re-pattern` appropriately for a literal path vs. a regex.  See https://github.com/puppetlabs/trapperkeeper-authorization/blob/1ac74458d6ee857f0fdcbf198584f635ed0ed1bd/src/puppetlabs/trapperkeeper/authorization/rules.clj#L67-L85.

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
camlow325, line 40: Also, the method conversion would happen "for free", I think, if we were layering on top of `build-rule`.

test/puppetlabs/trapperkeeper/services/authorization/authorization_service_test.clj
```diff
@@ -0,0 +1,58 @@
...
+
+(def minimal-config
+  "Minimal config required to bootstrap AuthorizationService with
+  EchoReverseService as a dependant consumer.
+
+  An empty list representing no rules with an expected behavior of
+  default-deny originating in the authorization library itself."
+  {:authorization {:rules []}})
+
+(def base-request
```
camlow325, line 38: Could use `ring.mock.request/request` to create the request map.

test/puppetlabs/trapperkeeper/services/authorization/authorization_service_test.clj
```diff
@@ -0,0 +1,58 @@
...
+  "A basic request to feed into the tests"
+  {:uri "/foo/bar"
+   :request-method :get
+   :remote-addr "127.0.0.1"})
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Tests
+
+(deftest ^:integration wrap-handler-test
```
camlow325, line 47: I think this is a good test and understand that we're not looking to retest the functionality of the rules translation or `wrap-authorization-check` middleware layer at this level.  It does, however, feel like we could use a couple of more tests here to at least convince ourselves that when a request is handled that the middleware is authorizing against some non-empty rule content from the configuration.  Maybe could just add a couple of extra tests with a single rule where the request being made is denied or allowed as expected?  If there's a good way to mock around the middleware function to avoid having to generate an SSL cert to make the test more system-level like, I'd be fine with that.

project.clj
jeffmccune: I don't know?

project.clj
jeffmccune: I use it quite a bit, but agree it doesn't need to be in the project.clj  Should I remove it?

src/puppetlabs/trapperkeeper/services/authorization/authorization_service.clj
jeffmccune: Addressed in f98bb15

src/puppetlabs/trapperkeeper/services/authorization/authorization_service.clj
jeffmccune: > Could just put it under ":rules" at the top-level.

Addressed in c10fe62

src/puppetlabs/trapperkeeper/services/authorization/authorization_service.clj
jeffmccune: Addressed in 21b7c83

project.clj
nwolfe: Also, FYI there's KS 1.2.0

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
jeffmccune, line 43: Should we change this validation function in this PR?  I'm in agreement that returning a collection of errors is the closest to ideal way to proceed, but this function is also going to necessarily change as we implement the full spec.

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
jeffmccune: Generally speaking, when it doesn't matter I prefer the system to tolerate mixed-case expressed by the user and normalize it internally, but that's just me.

I'll change this over to the clojure function instead of the interop method and assume these comments are addressed.  If that's insufficient please let me know what else is required before merge.

project.clj
nwolfe: Yes please.  Good to hear you use it, we'll know if we actually _can_ get rid of it then!  (We could the last time I tried a few months ago)

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
jeffmccune: `.toLowerCase` removed in f98bb15..83404d8

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
camlow325, line 43: Collection of errors sounds like a good idea to me, too.  I'd vote for doing it later on, though.  (If we do end up switching over to schema, that would also significantly change the way we'd batch up multiple violations as well).

project.clj
jeffmccune: Addressed in 83404d8..992a274

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
camlow325: I was more bothered by the code that ends up having to do `lower-case` in two different places than the fact that mixed case was being tolerated, I guess.  But I don't think it's a big deal either way, so I won't press on this further.

project.clj
jeffmccune: Addressed, updated to 0.8.1 in 992a274..ca213f9

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
jeffmccune: Yeah, that duplication bothers me too...

project.clj
jeffmccune: Addressed, removed spyscope in ca213f9..72a27ed

camlow325:
:chart_with_upwards_trend: Done reviewing for now.  Things are basically looking good to me.  The main things I'm interested in following-up on are:
- Would like to see if we can reuse more of the existing tk-auth code in constructing the ACL, like using what `config->rule` in config.clj does but just starting from a TK config representation instead of the HOCON one.  Maybe there's some intermediate massaging that would need to be done to get to that?
- Support for translating `method` from the config into the rule.  Maybe addressing the first bullet could implicitly take care of this point?
- Maybe add a couple of extra service tests - just enough to give more assurances that the rule representation from the config is what is actually being used during authorization checks for a request.

I'm open to moving any of this feedback that seems like it's going to take a lot of work to address off to a new ticket / PR if not getting the rest of this in is going to block other work which could otherwise get underway sooner.

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
jeffmccune: Addressed in 72a27ed..fa18768

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
jeffmccune: Good catch, thanks.  I didn't feel great about that anonymous function but I liked that I could see the positional arguments and so left it.  Wondered if I could eliminate it though, good to learn.

jeffmccune: Addressed in fa18768..29dce41

test/puppetlabs/trapperkeeper/services/authorization/authorization_service_test.clj
jeffmccune, line 38: That's what I was looking for and couldn't find.  It always trips me up that response is part of the ring API but request is instead tucked under the mock namespace.

test/puppetlabs/trapperkeeper/services/authorization/authorization_service_test.clj
jeffmccune, line 47: > Maybe could just add a couple of extra tests with a single rule where the request being made is denied or allowed as expected?

Agreed, I'll look into adding these.

jeffmccune:
> I'm open to moving any of this feedback that seems like it's going to take a lot of work to address off to a new ticket / PR if not getting the rest of this in is going to block other work which could otherwise get underway sooner.

@camlow325 With me being out on PTO Friday I don't think I can address all of these in this PR before then.  What I suggest is merging this PR as-is, but keeping the TK-258 ticket open or maybe even moving back to in-progress.  This will allow other work to continue while I fix-up the remaining comments in a second PR.

camlow325:
:+1: Going to merge this to unblock other work for now.  Will move the roll-up comments that weren't addressed yet back to the JIRA ticket.

PR #5, opened 2015-08-27
(TK-260) Add allow_unauthenticated support to authorization rules

jeffmccune:
This is a work in progress implementing [TK-260](https://tickets.puppetlabs.com/browse/TK-260)

### DIFF
--- a/src/puppetlabs/trapperkeeper/authorization/rules.clj
+++ b/src/puppetlabs/trapperkeeper/authorization/rules.clj
@@ -180,10 +180,10 @@
    request :- ring/Request
    name :- schema/Str]
   (if-let [ { matched-rule :rule matches :matches } (some #(match? % request) rules)]
-    (let [allow_unauthenticated? (true? (:allow_unauthenticated matched-rule))
-          authenticated? (true? (get-in request ring/is-authentic-key))]
-      (if (and (or allow_unauthenticated? authenticated?)   ; TK-260
-               (acl/allowed? (:acl matched-rule) name (:remote-addr request) matches))
+    (if (true? (:allow_unauthenticated matched-rule))
+      {:authorized true :message "allow_unauthenticated is true - allowed"}
+      (if (and (true? (get-in request ring/is-authentic-key)) ; authenticated?
+            (acl/allowed? (:acl matched-rule) name (:remote-addr request) matches))
         {:authorized true :message ""}
         {:authorized false :message (request->description request name matched-rule)}))
     {:authorized false :message "global deny all - no rules matched"}))
--- a/src/puppetlabs/trapperkeeper/authorization/rules.clj
+++ b/src/puppetlabs/trapperkeeper/authorization/rules.clj
@@ -15,6 +15,7 @@
    :type Type
    :path Pattern
    :method Method
+   (schema/optional-key :allow_unauthenticated) schema/Bool
    :acl acl/ACL
    (schema/optional-key :query-params) {schema/Str #{schema/Str}}
    (schema/optional-key :file) schema/Str
@@ -45,6 +46,15 @@
     pattern :- Pattern
     method :- Method]
     {:type type :path pattern :acl acl/empty-acl :method method}))
+
+(schema/defn equals-rule :- schema/Bool
+  "Test if two given rule are fully equal"
+  [a :- Rule b :- Rule]
+  (and
+    (= (:type a) (:type b))
+    (= (:acl a) (:acl b))
+    (= (:method a) (:method b))
+    (= (str (:path a)) (str (:path b)))))
 
 (schema/defn tag-rule :- Rule
   "Tag a rule with a file/line - useful for instance when the rule has been read
@@ -170,10 +180,12 @@
    request :- ring/Request
    name :- schema/Str]
   (if-let [ { matched-rule :rule matches :matches } (some #(match? % request) rules)]
-    (if (and (true? (get-in request ring/is-authentic-key))
-             (acl/allowed? (:acl matched-rule) name (:remote-addr request) matches))
-      {:authorized true :message ""}
-      {:authorized false :message (request->description request name matched-rule)})
+    (let [allow_unauthenticated? (true? (:allow_unauthenticated matched-rule))
+          authenticated? (true? (get-in request ring/is-authentic-key))]
+      (if (and (or allow_unauthenticated? authenticated?)   ; TK-260
+               (acl/allowed? (:acl matched-rule) name (:remote-addr request) matches))
+        {:authorized true :message ""}
+        {:authorized false :message (request->description request name matched-rule)}))
     {:authorized false :message "global deny all - no rules matched"}))
 
 (schema/defn authorized? :- schema/Bool
--- a/src/puppetlabs/trapperkeeper/authorization/rules.clj
+++ b/src/puppetlabs/trapperkeeper/authorization/rules.clj
@@ -47,15 +47,6 @@
     method :- Method]
     {:type type :path pattern :acl acl/empty-acl :method method}))
 
-(schema/defn equals-rule :- schema/Bool
-  "Test if two given rule are fully equal"
-  [a :- Rule b :- Rule]
-  (and
-    (= (:type a) (:type b))
-    (= (:acl a) (:acl b))
-    (= (:method a) (:method b))
-    (= (str (:path a)) (str (:path b)))))
-
 (schema/defn tag-rule :- Rule
   "Tag a rule with a file/line - useful for instance when the rule has been read
   from an authorization file."
--- a/src/puppetlabs/trapperkeeper/authorization/rules.clj
+++ b/src/puppetlabs/trapperkeeper/authorization/rules.clj
@@ -15,7 +15,7 @@
    :type Type
    :path Pattern
    :method Method
-   (schema/optional-key :allow_unauthenticated) schema/Bool
+   (schema/optional-key :allow-unauthenticated) schema/Bool
    :acl acl/ACL
    (schema/optional-key :query-params) {schema/Str #{schema/Str}}
    (schema/optional-key :file) schema/Str
@@ -180,8 +180,8 @@
    request :- ring/Request
    name :- schema/Str]
   (if-let [ { matched-rule :rule matches :matches } (some #(match? % request) rules)]
-    (if (true? (:allow_unauthenticated matched-rule))
-      {:authorized true :message "allow_unauthenticated is true - allowed"}
+    (if (true? (:allow-unauthenticated matched-rule))
+      {:authorized true :message "allow-unauthenticated is true - allowed"}
       (if (and (true? (get-in request ring/is-authentic-key)) ; authenticated?
             (acl/allowed? (:acl matched-rule) name (:remote-addr request) matches))
         {:authorized true :message ""}
--- a/src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
@@ -14,6 +14,10 @@
   {:allow #(rules/allow %1 %2)
    :deny #(rules/deny %1 %2)})
 
+(def new-rule-func-map
+  "This is a function map to allow programmatic execution of path/regex rules"
+  {:path rules/new-path-rule :regex rules/new-regex-rule})
+
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Private
 
@@ -30,10 +34,12 @@
   "Build a new Rule based on the provided config-map"
   [config-map]
   (let [rule-type (keyword (get-in config-map [:match-request :type] :path))
-        path (get-in config-map [:match-request :path])]
-    (if (= rule-type :path)
-      (-> (rules/new-path-rule path (method config-map)))
-      (-> (rules/new-regex-rule path (method config-map))))))
+        path (get-in config-map [:match-request :path])
+        rule-fn (rule-type new-rule-func-map)
+        rule (rule-fn path (method config-map))]
+    (if (true? (:allow_unauthenticated config-map))
+      (assoc rule :allow_unauthenticated true)
+      rule)))
 
 (defn- add-individual-acl
   "Add an individual acl to a given rule:
--- a/src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
@@ -9,6 +9,14 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Constants
 
+(def required-keys
+  "Keys required in an auth rule map."
+  [:path :type])
+
+(def required-or-key
+  "At least one of these keys is required in an auth rule map."
+  #{:deny :allow :allow-unauthenticated})
+
 (def acl-func-map
   "This is a function map to allow a programmatic execution of allow/deny directives"
   {:allow #(rules/allow %1 %2)
@@ -37,8 +45,8 @@
         path (get-in config-map [:match-request :path])
         rule-fn (rule-type new-rule-func-map)
         rule (rule-fn path (method config-map))]
-    (if (true? (:allow_unauthenticated config-map))
-      (assoc rule :allow_unauthenticated true)
+    (if (true? (:allow-unauthenticated config-map))
+      (assoc rule :allow-unauthenticated true)
       rule)))
 
 (defn- add-individual-acl
@@ -87,11 +95,11 @@
       (throw (IllegalArgumentException.
               (str "The authorization rule specified as " (pprint-rule rule)
                    " does not contain a '" (name k) "' key.")))))
-  (if (:allow_unauthenticated rule)
+  (if (:allow-unauthenticated rule)
     (if (some #{:deny :allow} (keys rule))
       (throw (IllegalArgumentException.
                (str "Authorization rule specified as  " (pprint-rule rule)
-                 " cannot have allow or deny if allow_unauthenticated."))))
+                 " cannot have allow or deny if allow-unauthenticated."))))
     (when-not (some #{:deny :allow} (keys rule))
       (throw (IllegalArgumentException.
                (str "Authorization rule specified as " (pprint-rule rule)
--- a/src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
@@ -87,10 +87,15 @@
       (throw (IllegalArgumentException.
               (str "The authorization rule specified as " (pprint-rule rule)
                    " does not contain a '" (name k) "' key.")))))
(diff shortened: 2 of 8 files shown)


src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -37,10 +41,12 @@
...
   [config-map]
-  (let [rule-type (keyword (get config-map :type :path))]
-    (if (= rule-type :path)
-      (-> (rules/new-path-rule (config-map :path) (method config-map)))
-      (-> (rules/new-regex-rule (config-map :path) (method config-map))))))
+  (let [rule-type (keyword (get config-map :type :path))
+        rule-fn (rule-type new-rule-func-map)
+        rule (rule-fn (config-map :path) (method config-map))]
+    (if (true? (:allow_unauthenticated config-map))
+      (assoc rule :allow_unauthenticated true)
```
jeffmccune: This feels a bit gross and smells bad to me.  Suggestions welcome to improve the flow here.

Maybe get rid of the conditional with something like?

``` clj
(->>
  (true? (:allow_unauthenticated config-map)
  (assoc rule :allow_unauthenticated))
```

src/puppetlabs/trapperkeeper/authorization/rules.clj
```diff
@@ -15,6 +15,7 @@
    :type (schema/enum :string :regex)
    :path Pattern
    :method (schema/enum :get :post :put :delete :head :any)
+   (schema/optional-key :allow_unauthenticated) schema/Bool
```
nwolfe: What do you think about adding an "invalid config" test for this new field to ensure that it's of the proper type?

See: https://github.com/puppetlabs/trapperkeeper-authorization/blob/master/test/puppetlabs/trapperkeeper/services/authorization/authorization_core_test.clj#L47

Or do we have near-term plans to change how validation is done? Can't remember, but I ended up writing these tests for the query-param stuff.

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -37,10 +41,12 @@
 (defn- build-rule
   "Build a new Rule based on the provided config-map"
   [config-map]
-  (let [rule-type (keyword (get config-map :type :path))]
-    (if (= rule-type :path)
-      (-> (rules/new-path-rule (config-map :path) (method config-map)))
-      (-> (rules/new-regex-rule (config-map :path) (method config-map))))))
+  (let [rule-type (keyword (get config-map :type :path))
+        rule-fn (rule-type new-rule-func-map)
+        rule (rule-fn (config-map :path) (method config-map))]
```
nwolfe: Admittedly I haven't spent much mental energy thinking about this, but...what are you thinking is the advantage of introducing function maps in this code?

I'm not against it, just curious.

jeffmccune: @nwolfe The advantage is that it eliminates the conditional check, which seemed pretty awkward to me.

jeffmccune: @nwolfe To clarify a bit more, in my mind the function has two different behaviors with the condition in place, with it removed the function has only one behavior, so it's easier for me to think about.  That's what I see as the advantage, at least personally.

nwolfe: Hm, I find this much harder to follow using a function map.  Considering that we don't expect there to be more than 2 or 3 different branches, I would prefer a straightforward `cond` expression.

> the function has two different behaviors with the condition in place

You could argue it still has different behaviors with a function map, it's just decoupled from the different types of rules that determine the behavior.  
I would consider this as a good thing if we knew we were going to be adding more types of rules and all we'd need to do is add a new function to the map, but I don't think that's the case.
Or if we had multiple places where we wanted to use the function map and reduce duplication.

I know I know, higher-order functions and all, but I think a `cond` is easier in this case.

Not going to make a fuss over it though, just explaining my thinking :smile: It's a fairly isolated piece of code, so not a big deal.

@camlow325 you got any thoughts?

src/puppetlabs/trapperkeeper/authorization/rules.clj
camlow325: For all of the settings that we've been adding to our Clojure projects (Trapperkeeper / Puppet Server), I believe we've been pretty consistent about using a hyphen instead of an underscore in the name.  Would you be okay with changing the references here from `allow_unauthenticated` to `allow-unauthenticated` - both where used as a keyword in the code and as the name of the key represented in the HOCON file loaded from disk?

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
camlow325: I actually kind of prefer what @jeffmccune did here over what's currently in the code.  Less conditional logic and enforces consistency in the `rule-fn` signature.  MIght make this easier to maintain in the future if we were to add more `rule-fn` types as well.  I don't have a strong opinion on this, though, so if @nwolfe were to convince @jeffmccune to forgo this change, I'd be okay with that, too.

camlow325:
I wrote a small unit test against the code on the branch using `:allow_unauthenticated true`.  That test failed for me.  Here's the code:

``` clj
(defservice hello-service-using-tk-authz
            [[:AuthorizationService wrap-with-authorization-check]
             [:WebserverService add-ring-handler]]
            (init [this context]
                  (add-ring-handler
                    (wrap-with-authorization-check
                      (fn [_]
                        {:status  200
                         :headers {"Content-Type" "text/plain"}
                         :body    "Hello, World!"}))
                    "/hello")
                  context))

(testing "allow unauth with no client cert successful"
    (with-test-logging
      (with-app-with-config
        app
        [jetty9-service hello-service-using-tk-authz authorization-service]
        (-> minimal-config
            (assoc-in [:authorization :rules]
                      [{:path                  "/hello/world"
                        :type                  "path"
                        :method                :get
                        :allow_unauthenticated true
                        :allow                 ["*"]}])
            (assoc :webserver {:client-auth "none"
                               :ssl-host    "localhost"
                               :ssl-port    8140
                               :ssl-ca-cert ".../ca-cert.pem"
                               :ssl-cert    ".../localhost-cert.pem"
                               :ssl-key     ".../localhost-privkey.pem"}))
        (let [response (http-client/get "https://localhost:8140/hello/world"
                                        {:as          :text
                                         :ssl-ca-cert ".../ca-cert.pem"})]
          (is (= 200 (:status response))
              (ks/pprint-to-string response))))))
```

Note that the locations of the certificates and keys were just some valid ones I had around outside of the trapperkeeper-authorization repo.  For the HTTP request, rather than the expected "200" status code, instead the test gets a "500" status code.

The status payload contained the following text:

```
HTTP ERROR: 500</h2>\n<p>Problem accessing /hello/world. Reason:\n<pre>    Input to add-authinfo does not match schema: [(named {:ssl-client-cert (not (instance? java.security.cert.X509Certificate nil))} request)]
```

I think we need to have the implementation tolerate a ring request with an `:ssl-client-cert` of nil rather than forbid it by schema.  In this case, if `allow_unauthenticated` is true for the matched rule, we'd need to treat the request as allowed.  Also, I think it would be good if we add at least _some_ tests into the project which do real HTTP requests so that we can catch stuff like this more easily in the future.

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
nwolfe: After talking IRL I'm good with these changes :high_brightness:

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
camlow325: Your revision seems fine to me.

test/puppetlabs/trapperkeeper/authorization/testutils.clj
```diff
@@ -2,9 +2,12 @@
   (:require [puppetlabs.ssl-utils.simple :as ssl]))
 
 (defn request
-  "Builds a ring request"
-  [path method certificate ip]
-  {:uri path :request-method method :remote-addr ip :ssl-client-cert certificate})
+  "Build a ring request for testing"
+  ([path method] (request path method "127.0.0.1"))
+  ([path method ip] {:uri path :request-method method :remote-addr ip})
```
camlow325: Might still be nice to use ring-mock/request if that's workable?

test/puppetlabs/trapperkeeper/services/authorization/authorization_service_test.clj
```diff
@@ -89,9 +137,22 @@
             {:keys [status body]} (app req)]
         (is (= status 200))
         (is (= body "Prefix: !dlroW olleH")))))
+  (testing "(TK-260) Authorizing unauthenticated requests"
```
camlow325, line 152: Might be good to throw another test in here somewhere which validates that that authenticated request - one that should also be allowed by rule - will still end up being allowed for a rule which has `allow_authenticated: true`.  I think that will work fine with the changes as you have them now.  The extra test might help ensure that we don't inadvertently break that when we make changes around this stuff later.

camlow325, line 152: Might also be good to add a test which has an unauthenticated request being made to a URI which does match a rule but where the rule has been configured with `allow_unauthenticated: false` (explicitly or implicitly by not having `allow_unauthenticated` specified) to confirm that the request is still denied.

test/puppetlabs/trapperkeeper/authorization/ring_middleware_test.clj
```diff
@@ -30,15 +30,15 @@
...
-      (let [response (ring-handler (request "/path/to/foo" :get test-domain-cert "127.0.0.1"))]
+      (let [response (ring-handler (request "/path/to/foo" :get "127.0.0.1" test-domain-cert))]
         (is (= 200 (:status response)))
         (is (= "hello" (:body response)))))
     (testing "access denied when cert CN is not in the rule"
-      (let [response (ring-handler (request "/path/to/foo" :get test-other-cert "127.0.0.1"))]
+      (let [response (ring-handler (request "/path/to/foo" :get "127.0.0.1" test-other-cert))]
         (is (= 403 (:status response)))
         (is (= "Forbidden request: www.other.org(127.0.0.1) access to /path/to/foo (method :get) (authentic: true)" (:body response)))))
     (testing "access denied when cert CN is explicitely denied in the rule"
```
camlow325: Not introduced by this PR but s/explicitely/explicitly.

camlow325:
:chart_with_upwards_trend: Done reviewing for now.  Most important comments to address, I think, are:
- Should handle `:ssl-client-cert nil` on the Ring request as "unauthenticated" for the purpose of authorizing a request since that's apparently what happens with a Jetty->Ring request over SSL in which the HTTP client has not included a certificate.
- No comment about this on the PR, but on the JIRA ticket I added a comment that it would be good to treat a rule that has `allow_unauthenticated: true` that also has either a `deny` or `allow` specification as "invalid".  @ahpook raised this suggestion in that he thought that combining these directives would not seem to make sense and most likely would have been a mistake unintentionally made by the user.  
- Rename `allow_unauthenticated` to `allow-unauthenticated` for consistency with other Clojure project settings.
- Could add a couple of extra `allow_unauthenticated` tests as mentioned in previous comments.

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
nwolfe: What if we had the `config->rule` always insert optional keys with their default values?  Presumably we don't have any situations where the lack of a key in the configuration means something different than it's default value.

I did this with the query-param support - the configuration file can have a single string or a list of strings, but once it's parsed by `config->rule` the in-memory rule will always have a list of strings (which may contain a single string).

I think this sort of general pattern might help in reducing this sort of conditional checking.  This part could be written as:

``` clj
(assoc rule :allow_unauthenticated (:allow_unauthenticated config-map)
```

...or a `->>`, whatever.

Granted, I haven't traced this all the way through the code to see if there'd be a problem with this approach in this case, but it's a valid consideration nonetheless.

@jeffmccune Would that be an improvement?

nwolfe: Whoops, I take some of that back already :smile: 

> What if we had the config->rule always insert optional keys with their default values? Presumably we don't have any situations where the lack of a key in the configuration means something different than it's default value.

The optional `:query-params` map might not fit into this pattern that well, actually.  So, maybe we can't just do this for everything as easily as I was thinking.

In the case of this `:allow_unauthenticated` field though the question/suggestion still applies though.

jeffmccune: > @jeffmccune Would that be an improvement?

I think so, I agree that having these transformation functions return a complete set of keys, even if the input configuration is lacking those keys, is an improvement.  It makes it much easier to reason about the output of these functions.

src/puppetlabs/trapperkeeper/authorization/rules.clj
jeffmccune: Sure, will change to hyphens.

jeffmccune:
@camlow325 

> I think we need to have the implementation tolerate a ring request with an :ssl-client-cert of nil rather than forbid it by schema.

Ah, so I ran into this issue as well.  It's weird because there are actually 3 states and I think the 3rd state is causing an issue with the prismatic schema validation.  The three states I see are:
1. `:ssl-client-cert` is absent from the map.
2. `:ssl-client-cert` is present in the map and has a value that is a certificate type.
3. `:ssl-client-cert` is present in the map and has a value of `nil`.

It's the 3rd case that's causing the schema validation issue.  I didn't commit it, but I tried to use `schema/maybe` and even that didn't work.

Basically, I think we need to try and reduce this down to only 2 possible states and try to completely eliminate the edge case of having the key bound to a nil value.  This, in turn, would prevent the code from having to deal with a nil value given the presence of the key in the map.

With that said, it may be much easier and faster to just have the code deal with a nil.  I'll try and work that up now.

camlow325:
@jeffmccune Yeah, I'm surprised the schema change didn't take care of it.  With that same test I mentioned above, I had just changed the reference to the `:ssl-client-cert` in the `Request` schema from ring.clj from...

``` clojure
(schema/optional-key :ssl-client-cert) X509Certificate
```

... to ...

``` clojure
(schema/optional-key :ssl-client-cert) (schema/maybe X509Certificate)
```

... and the test passed.  All of the other tests were still passing as well.  Not sure if that's the only change we'd need, but maybe...?

test/puppetlabs/trapperkeeper/services/authorization/authorization_service_test.clj
```diff
@@ -89,9 +137,22 @@
...
+        (is (= status 403))
+        (is (= body "global deny all - no rules matched")))
+      (let [req (assoc unauthenticated-request :uri "/puppet-ca/v1/certificate/ca"
+                                               :body "FOOBAR")
+            {:keys [status body]} (app req)]
+        (is (= status 200))
+        (is (= body "Prefix: RABOOF")))
+      (let [req (assoc unauthenticated-request :uri "/not/covered/by/rules")
+            {:keys [status body]} (app req)]
+        (is (= status 403)))))
```
camlow325: Seems like here we should either have an extra assertion for the value of `body` or just remove the `body` key from what's being destructured since, in the latter case, nothing would be using it.

jeffmccune:
> @jeffmccune Yeah, I'm surprised the schema change didn't take care of it.

It turned out to be an issue with stale state in the REPL.  Once I killed it and re-started it I observed the behavior you described.

src/puppetlabs/trapperkeeper/authorization/rules.clj
```diff
@@ -46,6 +47,15 @@
...
     {:type type :path pattern :acl acl/empty-acl :method method}))
 
+(schema/defn equals-rule :- schema/Bool
+  "Test if two given rule are fully equal"
+  [a :- Rule b :- Rule]
+  (and
+    (= (:type a) (:type b))
+    (= (:acl a) (:acl b))
+    (= (:method a) (:method b))
+    (= (str (:path a)) (str (:path b)))))
```
nwolfe: Whoops, delete this function

jeffmccune:
EDIT: Still need to add some more tests.

@camlow325 @nwolfe OK, I think I addressed all of the comments JB mentioned in the following list.  Please let me know if anything is missing:
- Should handle :ssl-client-cert nil on the Ring request as "unauthenticated" for the purpose of authorizing a request since that's apparently what happens with a Jetty->Ring request over SSL in which the HTTP client has not included a certificate.
- No comment about this on the PR, but on the JIRA ticket I added a comment that it would be good to treat a rule that has allow_unauthenticated: true that also has either a deny or allow specification as "invalid". @ahpook raised this suggestion in that he thought that combining these directives would not seem to make sense and most likely would have been a mistake unintentionally made by the user.
  - Rename allow_unauthenticated to allow-unauthenticated for consistency with other Clojure project settings.
  - Could add a couple of extra allow_unauthenticated tests as mentioned in previous comments.

test/puppetlabs/trapperkeeper/authorization/testutils.clj
jeffmccune: Yeah...  Sorry I didn't do this.  Do you think it would be OK to keep this here and use ring-mock-request to build the initial map?

jeffmccune:
Actually, looks like I still need to add some more tests too.

test/puppetlabs/trapperkeeper/authorization/testutils.clj
camlow325: Yeah, I'd be fine with assoc'ing extra stuff in after using ring-mock/request to build the initial map.  @nwolfe - WDYT?

nwolfe: Yeah, although I thought we would use the various "add something to the request" helpers rather than hardcoding the keys and assoc'ing them in ourselves; that was the point of doing this, right?

camlow325: Yeah, where ring has helpers for adding stuff onto its base request map, I agree it definitely makes sense for us to use those.

camlow325:
:+1: @jeffmccune - we're just waiting on a couple of extra tests at this point from your side, right?

camlow325:
@nwolfe, @dankreek - have either of you done a full review on this one?  If not, could one of you?

dankreek:
@jeffmccune this needs a rebase

dankreek:
:+1:  LGTM, just waiting on a rebase for merge.

jeffmccune:
> @jeffmccune - we're just waiting on a couple of extra tests at this point from your side, right?

Yes, adding them in now.  Just finished the rebase.

jeffmccune:
@camlow325 @dankreek @nwolfe OK, I added some more tests and rebased this on top of master.  It should be good to merge now.

The only comment I did't really address is @camlow325's request for some end-to-end tests that use "real" HTTP requests instead of ring request maps.  I decided not to put this in because the issue that raised this concern is adequately covered by what we have at the ring layer.  Adding a full HTTP test wouldn't have caught the issue any better than what we have.

I'm happy to add some HTTP layer tests as a follow on PR, but I didn't think it was worth holding up this one any longer, especially as we move stuff around in the configuration structures.

camlow325:
:+1: @dankreek - whatcha think?

dankreek:
@camlow325 It's not too late... to merge it.. merge it good! :+1:

PR #6, opened 2015-08-31
(TK-259) Support matching on query parameters

nwolfe:


nwolfe:
@camlow325 @jeffmccune I wrote the configuration (file) parameter as the plural `:query-params` despite all of our existing communication that is using the singular `:query-param`.  

Seems like it should be plural to me, but not a big deal; it's very likely the singular form was deliberate :) Happy to make the change.

jeffmccune:
Agree with plural.  "query parameters rest" in Google returns 20M results with "query parameter rest" returning 13M.

camlow325:
:+1: on the idea of making it plural.  @nwolfe - when you get a chance, can you update the JIRA ticket and the RFC doc with `query-params` so we're using the same term everywhere?

nwolfe:
> @nwolfe - when you get a chance, can you update the JIRA ticket and the RFC doc with query-params so we're using the same term everywhere?

Done.

src/puppetlabs/trapperkeeper/authorization/rules.clj
```diff
@@ -64,6 +56,18 @@
...
+  "Add a query parameter matching value(s) to a rule. New values will be
+   appended to existing values.
+
+   The query parameters are in a map under the `:match-request` section of the
+   rule. Keys in the map are strings corresponding to the query parameters to
+   match, and the values are sequences of strings of acceptable values."
+  [rule :- Rule
+   param :- schema/Str
+   value :- (schema/either schema/Str [schema/Str])]
+  (update-in rule [:query-params param] into (flatten [value])))
```
camlow325: Since it looks like we end up converting the flattened value into a set each time we do a `query-params-match?` later, should we just do the conversion to set here?

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -56,6 +56,13 @@
   (->> (select-keys config-map #{:allow :allow-ip :deny :deny-ip})
        (reduce #(add-individual-acl (first %2) (second %2) %1) rule)))
 
+(defn add-query-params
+  "Add any query parameters specified in configuration to the rule."
+  [rule {:keys [query-params]}]
+  (if query-params
```
camlow325: Do we need this if statement?  Can we just do the `reduce-kv` unconditionally and have it effectively return back the original rule untouched in the case that `query-params` was not in the input map?

test/puppetlabs/trapperkeeper/authorization/rules_test.clj
```diff
@@ -68,6 +52,31 @@
       (testing (str "matching " x)
         (is (= (:rule (rules/match? rule (request "/path/to/resource" x))) rule))))))
 
+(deftest test-matching-query-parameters
```
camlow325, line 55: I think it would be good to have at least a few tests for equality where you have multiple query param keys, not just multiple values for one key, in the rule.

camlow325:
:+1: with just a few cosmetic comments.  Most significant one was suggesting that we might add a couple of more tests for rule matches against multiple query-param keys.

src/puppetlabs/trapperkeeper/authorization/rules.clj
```diff
@@ -119,11 +123,20 @@
   [a b]
   (or (= a b) (some #{:any} [a b])))
 
+(defn- query-params-match?
```
dankreek, line 126: I'm starting to see `defn-` come back in our codebase. Is this no longer considered a sub-optimal practice?

masterzen, line 126: It's certainly my fault. 
What's the underlying reason for not using `defn-`?

test/puppetlabs/trapperkeeper/authorization/rules_test.clj
nwolfe, line 55: I did that in the integration test, but I'll look at adding one here too I guess

src/puppetlabs/trapperkeeper/authorization/rules.clj
nwolfe: Yeah, I had that fleeting idea too - I'll try it out

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
nwolfe: Maybe. The `query-params` here is straight from the config, so it's possible it will be `nil`.  I think the `rules/query-param` will gladly add empty/nil values, which we don't want, so I put the conditional here.

I'll play around with it a bit to see if we can improve it.

nwolfe: Ohhh I see, yes actually I think that'll work as `reduce-kv` will just return `rule` if the `coll` is empty or nil. Nice catch.

src/puppetlabs/trapperkeeper/authorization/rules.clj
nwolfe: Ugh this line doesn't behave the way we want if we try to make them sets here.  I thought this was just a matter of wrapping a `set` call around `flatten`, but it's not.

I'll continue playing with it

nwolfe:
@camlow325 Feedback addressed.

test/puppetlabs/trapperkeeper/authorization/rules_test.clj
```diff
@@ -68,6 +52,38 @@
...
+  (let [rule (rules/new-path-rule "/path/to/resource" :any)
+        foo-rule (rules/query-param rule "environment" "foo")
+        foo-bar-rule (rules/query-param rule "environment" ["foo" "bar"])
+        multiples-rule (-> rule
+                           (rules/query-param "beatles" ["lennon" "starr"])
+                           (rules/query-param "monkees" "davy"))]
```
camlow325, line 61: s/davy/mickey

dankreek, line 61: Peter was clearly the superior Monkey

camlow325:
:+1:

src/puppetlabs/trapperkeeper/authorization/rules.clj
camlow325, line 126: I think the rationale for not using `defn-` was that it makes it harder to write unit tests against those functions when your testing namespace is outside of the code's namespace.  To avoid that, we've instead frequently decorated functions intended to be "private" with a comment pragma like this:

https://github.com/puppetlabs/trapperkeeper/blob/trapperkeeper-1.1.0/src/puppetlabs/trapperkeeper/config.clj#L40-L53

I don't think we've always been consistent about that, though.  For example, we still have a number of `defn-` references in the trapperkeeper project and other PL Clojure projects like clj-http-client.

@KevinCorcoran - do you know what our "state of the art" practice for this is considered to be?

src/puppetlabs/trapperkeeper/authorization/rules.clj
```diff
@@ -64,6 +56,18 @@
       (assoc :file file)
       (assoc :line line)))
 
+(schema/defn query-param :- Rule
+  "Add a query parameter matching value(s) to a rule. New values will be
+   appended to existing values.
+
+   The query parameters are in a map under the `:match-request` section of the
```
nwolfe: Whoa, wrong key here...lemme update this

src/puppetlabs/trapperkeeper/authorization/rules.clj
masterzen, line 126: Thanks for the explanation, that makes sense. One could argue that you can test through the public interface of the module and your tests shouldn't have any knowledge of the gory internal details (hence `defn-` usage), but I see the point.

README.md
```diff
@@ -80,28 +80,37 @@ ordering _allows_ before _deny_, and with an implicit _deny all_.
 A `Rule` is:
 * a path or a regex
 * an optional method (get, post, put, delete, head)
+* an optional map of request query parameters
 * an ACL
 
 Using the internal DSL to build a rule is very simple:
```
dankreek, line 86: But can it be metaprogrammed?

camlow325, line 86: And ... is the hidden singleton lurking behind it?  :smile: (https://github.com/puppetlabs/puppet/blob/4.2.0/lib/puppet/util/metaid.rb#L2)

src/puppetlabs/trapperkeeper/authorization/rules.clj
KevinCorcoran, line 126: @camlow325 - I don't know if we have an accepted best practice around `defn-` right now.  I know we've waffled a bit on some projects (as I have personally).  And my observation has been that the Clojure community at-large does not exactly have its mind made up either - I see it used heavily in some projects and then not at all in others.

dankreek, line 126: I tend to use "private" methods to bootstrap the middle of a function call chain for testing. It's possible this may not be the best way to test something, but it's how I tend to roll. In general, when making a library I would generally only expose "public" functions from a core namespace, and stuff all the internals somewhere else. In this case, the "API" level stuff is in the service definition.

dankreek:
Welp, LGTM, :+1:

PR #7, opened 2015-09-03
(TK-275) Add :match-request to config rule

nwolfe:
This commit adds a new section to the rule specified in the
configuration file called :match-request that now contains the :type,
:path, :method, and :query-params fields.

Note this :match-request section only exists in the configuration file.
In-memory Rules are not changed; the keys will be at the top-level.

### DIFF
--- a/src/puppetlabs/trapperkeeper/authorization/rules.clj
+++ b/src/puppetlabs/trapperkeeper/authorization/rules.clj
@@ -12,9 +12,9 @@
 (def Rule
   "An ACL rule, with no less than a matching path, possibly a method list and an acl"
   {
-   :type (schema/enum :string :regex)
+   :type Type
    :path Pattern
-   :method (schema/enum :get :post :put :delete :head :any)
+   :method Method
    :acl acl/ACL
    (schema/optional-key :query-params) {schema/Str #{schema/Str}}
    (schema/optional-key :file) schema/Str
@@ -78,7 +78,7 @@
   ([path :- schema/Str]
     (new-path-rule path :any))
   ([path :- schema/Str
-   method :- Method]
+    method :- Method]
     (new-rule :string (path->pattern path) method)))
 
 (schema/defn new-regex-rule :- Rule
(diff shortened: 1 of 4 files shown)


src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -9,14 +9,6 @@
...
 ;;; Constants
 
-(def required-keys
-  "Keys required in an auth rule map."
-  [:path :type])
-
-(def required-or-key
-  "At least one of these keys is required in an auth rule map."
-  #{:deny :allow})
-
```
nwolfe, line 19: I inlined these in the `validate-auth-config-rule!` function because the exception message(s) contains hardcodings for these keys anyway, so I don't see any value in pulling these out.

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -58,19 +51,19 @@
...
+
+   This assumes the configuration has been validated via
+   `validate-auth-config-rule!`."
   [m]
-  (if (contains? m :path)
-    (-> m
-        build-rule
-        (add-acl m)
-        (add-query-params m))
-    (throw (Exception. "Invalid config - missing required `path` key"))))
```
nwolfe, line 73: We should be doing all validation in `validate-auth-config-rule!`, so this doesn't make sense here. I added note on the docstring to mention this.

test/puppetlabs/trapperkeeper/services/authorization/authorization_core_test.clj
```diff
@@ -133,7 +149,8 @@
         (is (= (str path) "^\\Q/foo/bar/baz\\E")))
       (is (= :any method))))
   (testing "Given a basic allow rule with a specific :put method"
-    (let [m (merge base-path-auth allow-list {:method :put})
+    (let [m (merge-with merge base-path-auth allow-list
+                        {:match-request {:method :put}})
```
nwolfe, line 153: :hand: :hand: :gun:

src/puppetlabs/trapperkeeper/authorization/rules.clj
```diff
@@ -12,9 +12,9 @@
 (def Rule
   "An ACL rule, with no less than a matching path, possibly a method list and an acl"
   {
-   :type (schema/enum :string :regex)
+   :type Type
```
camlow325, line 15: Good call.

camlow325:
:+1: @dankreek - WDYT?

dankreek:
@camlow325 assuming you already did a functional review on this, it LGTM. :+1:

PR #8, opened 2015-09-03
(TK-272) Fix deny "*" rules

dankreek:


src/puppetlabs/trapperkeeper/authorization/acl.clj
```diff
@@ -79,30 +76,73 @@
...
+                :type :allow-all
+                :qualifier :exact
+                :length nil
+                :pattern []})
+
+(def deny-all {:auth-type :deny
+               :type :regex
+               :qualifier :exact
+               :length nil
+               :pattern #"^/.*/$"})
```
camlow325: I don't think this pattern is correct.  I think we want `#"^.*$"` or just `#"^*$"`.  With what you have, the `match-name?` function will return `false` when checking to see if this ACE matches a request name except for cases where the name is delimited with a forward-slash and back-slash.  So `/blah/` or just `//` would return `true` for `match-name?` but `blah` would return `false`.  For a wildcard like this, I think we want `match-name?` to return `true` regardless of the value in the name.

To validate this in a unit test, for example, try adding an allow ACE to the rule being used the new "Denied when deny all" test and note that if the request would end up being allowed if the name were delimited with a forward slash and back slash but denied otherwise.

src/puppetlabs/trapperkeeper/authorization/acl.clj
```diff
@@ -79,30 +76,73 @@
...
    pattern :- schema/Str]
   (cond
-    ; global allow
-    (= "*" pattern) allow-all
+    (and (= "*" pattern) (= type :allow))
+    allow-all
+
+    ; global deny
+    (and (= "*" pattern) (= type :deny))
+    deny-all
```
camlow325: The subtle differences between how `allow-all` and `deny-all` are handled in the code may look a little strange here.  It seems like maybe the significance of `allow-all` as a first class "type" is basically so that the `ace-compare` logic can weight an `allow-all` ACE as "more important" than any other ACE.  Arguably, it would make sense to weight a `deny-all` ACE as "more important" than any other ACE - with the possible exception of `allow-all`.

That being said, though, I think we might end up simplifying the ACE weighting to just weight any `deny` ACEs above any `allow` ACEs as an outcome of the work to be done for https://tickets.puppetlabs.com/browse/TK-268.  So given that, I think I'm good with just keeping the notion of `deny-all` self-contained here.

Maybe you could inline the `deny-all` map here rather than creating the separate `def` that only gets used once, but that feels cosmetic.  I'm good either way on that part.

camlow325:
Beyond the one change to the pattern used in the `deny-all` map, I'm :+1: on this.

src/puppetlabs/trapperkeeper/authorization/acl.clj
dankreek: Yeah, this was my thinking exactly. I didn't want to spend a whole bucket of time on straightening out the logic in this PR, since we're going to revisit it later. But inlining wouldn't be a bad idea.

src/puppetlabs/trapperkeeper/authorization/acl.clj
```diff
@@ -79,30 +76,69 @@
   [pattern]
   (-> pattern (str/lower-case) (str/split #"\.") reverse vec))
 
-(def allow-all { :auth-type :allow :type :allow-all :qualifier :exact :length nil :pattern [] })
```
nwolfe, line 82: Can't get rid of this because https://github.com/puppetlabs/trapperkeeper-authorization/blob/master/src/puppetlabs/trapperkeeper/authorization/acl.clj#L129

nwolfe:
:+1:

PR #9, opened 2015-09-06
(TK-277) Avoid slurping request body

camlow325:
This commit changes tk-authz to use a custom middleware function which
parses the request query string into request parameters without, in the
case of a urlencoded form post, slurping the request body into a string
which is then not available to any downstream consumers of the Ring
request map.  The previous commit would slurp the request body in this
case because this is the default behavior performed by Ring's
`wrap-params` middleware function, which tk-authz was previously using.

### DIFF
--- a/src/puppetlabs/trapperkeeper/authorization/ring_middleware.clj
+++ b/src/puppetlabs/trapperkeeper/authorization/ring_middleware.clj
@@ -1,5 +1,7 @@
 (ns puppetlabs.trapperkeeper.authorization.ring-middleware
   (:require [schema.core :as schema]
+            [ring.middleware.params :as ring-params]
+            [ring.util.request :as ring-request]
             [puppetlabs.trapperkeeper.authorization.rules :as rules]
             [puppetlabs.trapperkeeper.authorization.ring :as ring]
             [puppetlabs.ssl-utils.core :as ssl-utils])
@@ -8,7 +10,6 @@
 (schema/defn request->name :- (schema/maybe schema/Str)
   "Return the identifying name of the request or nil"
   [request :- ring/Request]
-  ; TODO TK-260 Return a name even if there is no SSL client cert
   ; TODO SERVER-763 Support the name from the X headers
   (if-let [certificate (:ssl-client-cert request)]
     (ssl-utils/get-cn-from-x509-certificate certificate)))
@@ -34,3 +35,20 @@
       (if (true? authorized)
         (handler req)
         {:status 403 :body message}))))
+
+(defn- assoc-query-params
+  [request]
+  (let [encoding (or (ring-request/character-encoding request) "UTF-8")]
+    (if (:query-params request)
+      request
+      (ring-params/assoc-query-params request encoding))))
+
+(schema/defn wrap-query-params :- IFn
+  "A ring middleware for destructuring query params from the request. This is
+   similar to ring's wrap-params except that it only looks at query string and
+   not at form params in the request body for a urlencodedform post.  tk-authz
+   uses this so that it doesn't consume a request body before downstream
+   middleware has a chance to access it."
+  [handler]
+  (fn [request]
+    (handler (assoc-query-params request))))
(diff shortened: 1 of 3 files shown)


jeffmccune:
:+1:

src/puppetlabs/trapperkeeper/authorization/ring_middleware.clj
```diff
@@ -34,3 +36,20 @@
...
+      request
+      (ring-params/assoc-query-params request encoding))))
+
+(schema/defn wrap-query-params :- IFn
+  [handler]
+  (str "A ring middleware for destructuring query params from the request.  "
+       "This is similar to ring's wrap-params except that it only looks at "
+       "query string and not at form params in the request body for a "
+       "urlencoded form post.  tk-authz uses this so that it doesn't consume "
+       "a request body before downstream middleware has a chance to access it.")
```
nwolfe: Was this intended to be a docstring?

src/puppetlabs/trapperkeeper/services/authorization/authorization_service.clj
```diff
@@ -23,4 +23,4 @@
     (let [rules (get-in (service-context this) [:rules])]
       (-> handler
           (wrap-authorization-check rules)
-          ring/wrap-params))))
```
nwolfe, line 26: Can get rid of the `ring` import now

nwolfe:
Yep, good thinking  - @camlow325 just curious what prompted this work?

:+1:

src/puppetlabs/trapperkeeper/services/authorization/authorization_service.clj
camlow325, line 26: Gone in aa84bf7.

src/puppetlabs/trapperkeeper/authorization/ring_middleware.clj
camlow325: Sure did, thanks.  Moved it before the params in aa84bf7.

camlow325:
When I first updated Puppet Server to tk-authz 0.1.2 in https://github.com/puppetlabs/puppet-server/pull/687, I found that the `params` being passed down into core Ruby Puppet for an incoming request ended up being empty - e.g., not even the "environment" parameter was being populated.  I tracked that back to the pre-wrapping of the Ring params middleware being done in tk-authz.

The Ring params middleware will add any query parameters it finds in the query string to both the `query-params` and `params` elements in the Ring request map.  Puppet Server has some code that clears out `params` before calling into the Ring params middleware - https://github.com/puppetlabs/puppet-server/blob/puppet-server-2.1.1/src/clj/puppetlabs/services/request_handler/request_handler_core.clj#L99-L104.  This had been done previously in order to empty out any params from comidi that might conflict with the query-params/form-params for an incoming request.  With the pre-wrapping of the Ring params middleware in tk-authz, the `params` wouldn't get filled back in a second time because the `query-params` had already been present on the map - https://github.com/ring-clojure/ring/blob/1.4.0/ring-core/src/ring/middleware/params.clj#L43-L45.

I eventually resolved the problem in the Puppet Server PR by just merging the `query-params` and `form-params` from the Ring request back into `params` after calling the Ring params middleware.  That just felt like a safer way to fill in the `params` given the different possible ways that `params` could have been filled into the request by any upstream middleware.  While debugging the problem, though, I realized that the `body` was also being consumed early by the Ring params middleware and so wouldn't be available for the logic in Puppet Server which expects it to be present - basically so that the `body`, even for a urlencoded form post, is provided in string form to core Ruby Puppet code.

As I mused in https://tickets.puppetlabs.com/browse/SERVER-594, I'm not sure that this change is strictly necessary for Puppet Server to function properly.  At the Puppet Server layer we're basically keeping that behavior in place for backward compatibility with Rack/WEBrick request handling for now and may end up ripping it out later - e.g., if/when Rack/WEBrick support is dropped.  At that point, the changes in this PR wouldn't strictly be necessary for compatibility with Puppet Server.  I don't think these changes would hurt, though, even at that point.

PR #10, opened 2015-09-09
(TK-271) Allow list of methods in a rule

dankreek:


test/puppetlabs/trapperkeeper/services/authorization/authorization_service_test.clj
```diff
@@ -102,15 +102,10 @@
...
-    {:path "/puppet-ca/v1/certificate_request"
-     :method :get
-     :type "path"}
-    :allow-unauthenticated true}
-   {:match-request
-    {:path "/puppet-ca/v1/certificate_request"
-     :method :put
-     :type "path"}
-    :allow-unauthenticated true}])
+    {:path   "/puppet-ca/v1/certificate_request"
```
camlow325, line 104: I didn't see anything in the tests which exercise this directly, so I wrote the following:

``` clojure
(testing "certificate_request"
    (let [app (build-ring-handler default-rules)]
      (let [req (request "/puppet-ca/v1/certificate_request/ca" :head "127.0.0.1" test-domain-cert)
        (is (= status 403))
        (is (= body "global deny all - no rules matched")))
      (let [req (request "/puppet-ca/v1/certificate_request/ca" :get "127.0.0.1" test-domain-cert)
            {:keys [status]} (app req)]
        (is (= status 200)))
      (let [req (request "/puppet-ca/v1/certificate_request/ca" :put "127.0.0.1" test-domain-cert)
            {:keys [status]} (app req)]
        (is (= status 200)))))
```

Both the ":get" and ":put" tests fail for this.  I believe this is because the `method` private function in "authorization_core.clj" assumes that the original method is a string rather than tolerating the ability for it to be a vector.  I changed the implementation of that from:

``` clojure
(keyword (get-in config-map [:match-request :method] :any))
```

... to ...

``` clojure
(let [method-from-config (get-in config-map [:match-request :method] :any)]
    (if (vector? method-from-config)
      (into [] (map keyword method-from-config))
      (keyword method-from-config)))
```

With this, the tests pass.  There may be a better way to fix this, but, in any event, I think we need to increase the test coverage and ensure that we have a path for doing multiple method rules when rules are created via the `build-rule` path.

camlow325:
I noticed we don't have any logic in `validate-auth-config-rule!` in "authorization_core.clj" for "method" in a rule.  Seems like it would be good to add that.

test/puppetlabs/trapperkeeper/services/authorization/authorization_service_test.clj
dankreek, line 104: Copied-and-pasted. Everything worked great.

camlow325:
Did some functional testing of these changes with the "patch" that I'd added to the private `method` function and with an "auth.conf" and Puppet Server.  Things seemed to be working the way I'd expected they would.

:+1: other than the comments I'd left about the vector conversion in the private `method` function and adding some extra validation of `method` in the `validate-auth-config-rule!` function.

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -36,7 +36,10 @@
 (defn- method
   "Returns the method key of a given config map, or :any if none"
   [config-map]
-  (keyword (get-in config-map [:match-request :method] :any)))
+  (let [method-from-config (get-in config-map [:match-request :method] :any)]
+    (if (vector? method-from-config)
+      (into [] (map keyword method-from-config))
```
nwolfe: Would `mapv` work here?

dankreek: Oh yeah! I always forget about little ole' `mapv`

src/puppetlabs/trapperkeeper/authorization/rules.clj
```diff
@@ -7,20 +7,19 @@
 ;; Schemas
 
 (def Type (schema/enum :string :regex))
-(def Method (schema/enum :get :post :put :delete :head :any))
+(def Methods (schema/enum :get :post :put :delete :head :any))
+(def Method (schema/either Methods [Methods]))
```
nwolfe: These schema names seem backwards

dankreek: Yeah I guess my initial intention was that `Method` was the schema name for the key called `method`

nwolfe: Might just inline the schema into Rule though, and just refer to multiples as `[Method]` rather than a dedicated schema for it

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -118,8 +125,16 @@
   (when-not (string? (:path (:match-request rule)))
     (throw (IllegalArgumentException.
             (str "The path set in the authorization rule specified as "
-                 (pprint-rule (:path (:match-request rule))) " is invalid. "
+                 (pprint-rule rule) " is invalid. "
```
nwolfe, line 152: Did you mean to change this?

dankreek, line 152: Yeah, just displaying the path by itself isn't all the useful.

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -17,6 +17,10 @@
   "At least one of these keys is required in an auth rule map."
   #{:deny :allow :allow-unauthenticated})
 
+(def valid-methods
+  "HTTP methods which are allowed to be configured in a rule."
+  ["get" "post" "put" "delete" "head" "any"])
```
nwolfe: We should re-use the `Method` schema rather than defining the list twice

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -118,8 +125,16 @@
...
+                 (pprint-rule rule) " is invalid. "
                  "It should be a string."))))
+  (when (and (not (nil? (:method rule)))
+             (nil? (some #(= % (:method rule)) valid-methods)))
+    (throw (IllegalArgumentException.
+             (str "The method specified in the authorization rule specified as "
+                  (pprint-rule rule) " is invalid. "
+                  "It should be either a string or list of strings that is equal "
+                  "to one of the following methods: 'get' 'post' 'put' 'delete' "
+                  "'head' 'any'"))))
```
nwolfe: I don't think `any` is actually an option, it's just what we use internally when one isn't configured.

nwolfe:
:+1:

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
dankreek: Hmm.. that's a good question. @camlow325, what do you think of this?

nwolfe: ...at least that's the conclusion we reached during the Module meeting yesterday

camlow325: Yeah, I like the idea of keeping `any` as something we only manage internally to tk-authz.  The user can then get `any` behavior by just omitting `method` from a rule.  FWIW, that would seem to be consistent with how this was handled in legacy Ruby Puppet's auth.conf - https://docs.puppetlabs.com/puppet/latest/reference/config_file_auth.html#method.

src/puppetlabs/trapperkeeper/authorization/rules.clj
camlow325: I'd agree that the names seem backward.

>  just refer to multiples as [Method] rather than a dedicated schema for it

It would seem onerous to have to modify each of the locations where `method` is passed in as a parameter to do the `schema/either` allowing for one method or a list of methods.  I could see naming the schema that either takes one or a vector of them as `Methods` with `Method` as the schema that has the possible enum values.  I'm not too fussed about how we do this, though.

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
nwolfe: Gaul it's a stupid `schema/enum`

dankreek:
@camlow325 allighty, I think we're now able to hit that big green thing

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -118,8 +125,16 @@
   (when-not (string? (:path (:match-request rule)))
     (throw (IllegalArgumentException.
             (str "The path set in the authorization rule specified as "
-                 (pprint-rule (:path (:match-request rule))) " is invalid. "
+                 (pprint-rule rule) " is invalid. "
                  "It should be a string."))))
+  (when (and (not (nil? (:method rule)))
```
camlow325: The `:method` is actually part of the `match-request` with the rule - like `path` - and not a top-level member of the rule.  So I don't think this is validating what it needs to be.

Also, will this work if the value for `method` is a vector?  In that case, I think we'd want to be validating that each of the values in the vector is one of the `valid-methods`.

test/puppetlabs/trapperkeeper/services/authorization/authorization_core_test.clj
```diff
@@ -144,7 +144,15 @@
                                        {:path "/"
                                         :type "path"
                                         :query-params {"env" [:notastring]}}
-                                       :allow "*"})))))
+                                       :allow "*"})))
+
+    (is (thrown? IllegalArgumentException
```
camlow325: I think this should be a `thrown-with-msg?` rather than a `thrown?` so that the message content is evaluated.  Also, I think the expected body of the message should be `#".* It should be either a string or list of strings .*"` - note the absent "a" between "or" and "list".  Could also change the error message thrown to include the "a".  Either way works for me.

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
dankreek: (facepalm) you're right, it won't work for a vector. And I thought it was above the match-request because I was looking at the config example in the "Trapperkeeper Authorization Design Doc" and misread the level it was at.

dankreek:
Facepalm has been lifted

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -79,6 +86,31 @@
...
+  "Is the given method contained in the list of valid methods?"
+  [m]
+  (let [method (str/lower-case m)]
+    (boolean (some #(= method %) valid-methods))))
+
+(defn valid-method?
+  "Returns true if the given rule contains either a valid method, or no speicfied
+  method."
+  [rule]
+  (if (nil? (:match-request rule)) false
```
camlow325: We could probably skip this check, right -- since `get-in ...` would return `nil` if the `:match-request` itself were `nil`?

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -79,6 +86,31 @@
       (add-acl m)
       (add-query-params m)))
 
+(defn in-valid-methods?
+  "Is the given method contained in the list of valid methods?"
+  [m]
+  (let [method (str/lower-case m)]
```
camlow325: Hmmm, do we ever normalize the `method` from the config file to lowercase?  Seems like that would happen in the `method` private function above it were happening, but not sure that it is.

test/puppetlabs/trapperkeeper/services/authorization/authorization_service_test.clj
```diff
@@ -51,66 +51,60 @@
   [{:match-request
     {:path "/puppet/v3/catalog/([^/]+)"
      :type "regex"
-     :method :get}
+     :method "get"}
```
nwolfe, line 54: We're changing the type?

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -79,6 +86,31 @@
...
+    (let [method (get-in rule [:match-request :method])]
+      (cond
+        (nil? method)
+        true
+
+        (string? method)
+        (in-valid-methods? method)
+
+        (vector? method)
+        (every? #(in-valid-methods? %) method)
```
camlow325: Could do `(every? in-valid-methods? method)` here?

test/puppetlabs/trapperkeeper/services/authorization/authorization_core_test.clj
```diff
@@ -144,7 +144,37 @@
...
+                       :type "path"
+                       :method ["put" "get"]}
+                      :allow "*"}))))
+
+    (is (not (nil? (validate-auth-config-rule!
+                     {:match-request
+                             {:path "/"
+                              :type "path"
+                              :method ["put"]}
+                      :allow "*"}))))))
```
nwolfe: What are these testing?

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
nwolfe: Don't need the anon function

camlow325: @nwolfe (highfive)

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -79,6 +86,31 @@
       (add-acl m)
       (add-query-params m)))
 
+(defn in-valid-methods?
+  "Is the given method contained in the list of valid methods?"
+  [m]
+  (let [method (str/lower-case m)]
+    (boolean (some #(= method %) valid-methods))))
```
nwolfe: Let's make the `valid-methods` a set and then we can probably get rid of this helper function

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
dankreek: ugh.. yep

test/puppetlabs/trapperkeeper/services/authorization/authorization_service_test.clj
dankreek, line 54: I don't think hocon will convert a string to a keyword.

test/puppetlabs/trapperkeeper/services/authorization/authorization_core_test.clj
camlow325: I think it would be better to work the last two into the `valid-configs-pass` tests since they aren't invalid configs.

test/puppetlabs/trapperkeeper/services/authorization/authorization_service_test.clj
camlow325, line 54: Yeah, that sounds right to me.  Better to start with strings since that's what the raw input parsed from HOCON would look like.

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
dankreek: you got it

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
dankreek: Well my thinking was that a missing :method is valid, but a missing :match-request isn't. But a missing `:match-request` should have been caught by now. So I can just remove it.

camlow325: Yeah, true.  Looks like we do already have an explicit check for the existence of `:match-request` in `validate-auth-config-rule!` so should be good there already.

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
nwolfe: If keeping it a list/vector is preferable we can also do `(contains? valid-methods method)`

dankreek:
New and improved. Now with less stupid.

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -36,7 +40,10 @@
 (defn- method
   "Returns the method key of a given config map, or :any if none"
   [config-map]
-  (keyword (get-in config-map [:match-request :method] :any)))
+  (let [method-from-config (get-in config-map [:match-request :method] :any)]
```
camlow325: @dankreek, only other question I had on this PR is whether we should lower-case the method around here somewhere since we allow mixed case to pass initial validation and I'm not sure we normalize to lower-case anywhere else before we use the rule.

src/puppetlabs/trapperkeeper/authorization/rules.clj
```diff
@@ -119,10 +118,16 @@
...
-(defn- method-match?
-  "Return true if both mathod a and b match or if one is :any"
-  [a b]
-  (or (= a b) (some #{:any} [a b])))
+(schema/defn method-match?
+  "Return true if the provided method is equal to the value of `specified`. If
+  `specified` is a list of methods, return true if `method` is contained in
+  `specified`. If `specified` is set to :any then all methods will result in
+  true."
+  [method :- schema/Keyword
```
nwolfe, line 126: Should this be the `Method` schema?

camlow325, line 126: Ring might hypothetically use an HTTP method for a request that we're not covering in our schema - e.g, PATCH or some user-defined method - so probably best not to restrict it here?

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
dankreek: Oh yeah, that's a good idea. It's fixed now.

nwolfe:
:black_joker:

PR #11, opened 2015-09-11
(TK-266) Rule sort-order & name

nwolfe:


README.md
```diff
@@ -74,65 +74,25 @@ ordering _allows_ before _deny_, and with an implicit _deny all_.
...
-To build a set of rule:
-
-~~~clojure
-(-> rules/empty-rules
-    (rules/add-rule (-> (new-path-rule "/path/to/resource")
-                  (allow "*.domain.org")))
-    (rules/add-rule (-> (new-regex-rule "(this|that)-resource")
-                  (allow "$1.domain.org"))))
-~~~
-
```
nwolfe, line 130: Not sure these docs were adding much so rather than refactor the examples I just deleted it.

src/puppetlabs/trapperkeeper/authorization/rules.clj
```diff
@@ -39,13 +40,19 @@
...
+   sort-order :- schema/Int
+   name :- schema/Str]
+  {:type type
+   :path (condp = type
+           :path (re-pattern (str "^" (Pattern/quote path)))
+           :regex (re-pattern path))
+   :acl acl/empty-acl
+   :method method
+   :sort-order sort-order
+   :name name})
```
nwolfe: This is a pretty significant refactoring I made.  I eliminated the constructor functions `new-path-rule` and `new-regex-rule`, which only provided the string path -> regex pattern conversion.

Now there's only one constructor, with a single arity - `new-rule`

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -9,23 +10,11 @@
...
 ;;; Constants
 
-(def required-keys
-  "Keys required in an auth rule map."
-  [:path :type])
-
-(def required-or-key
-  "At least one of these keys is required in an auth rule map."
-  #{:deny :allow :allow-unauthenticated})
-
```
nwolfe, line 19: These were unused and ended up slipping back in during a rebase

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -33,18 +22,15 @@
   [rule]
   (str/trim (ks/pprint-to-string rule)))
 
-(defn- method
-  "Returns the method key of a given config map, or :any if none"
-  [config-map]
-  (keyword (get-in config-map [:match-request :method] :any)))
```
nwolfe: inlined below

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -9,23 +10,11 @@
...
-  #{:deny :allow :allow-unauthenticated})
-
 (def acl-func-map
   "This is a function map to allow a programmatic execution of allow/deny directives"
   {:allow #(rules/allow %1 %2)
    :deny #(rules/deny %1 %2)})
 
-(def new-rule-func-map
-  "This is a function map to allow programmatic execution of path/regex rules"
-  {:path rules/new-path-rule :regex rules/new-regex-rule})
```
nwolfe, line 31: Don't need this is there's only one constructor function now, `new-rule`

test/puppetlabs/trapperkeeper/services/authorization/authorization_core_test.clj
```diff
@@ -48,109 +51,139 @@
...
+          (is (= (:sort-order parsed) (:sort-order rule)))
+          (is (= (:name parsed) (:name rule)))
+          (is (false? (contains? parsed :match-request))))))))
+
+(def testrule
+  "A valid rule definition as it would be found in the configuration file."
+  {:name "name"
+   :sort-order 1
+   :allow "allow"
+   :match-request {:path "/path" :type "path"}})
```
nwolfe, line 69: I refactored all the tests below to do `assoc`,`dissoc` on this rather than repeating explicit maps

test/puppetlabs/trapperkeeper/services/authorization/authorization_core_test.clj
```diff
@@ -190,14 +233,4 @@
...
-    (testing "Given a rule config with query parameters"
-      (is (= {"env" #{"prod" "staging"}}
-             (-> (merge-with merge multi-query-param base-path-auth)
-                 config->rule
-                 :query-params)))
-      (testing "single values are converted to sets"
-        (is (= {"environment" #{"production"}}
-               (-> (merge-with merge single-query-param base-path-auth)
-                   config->rule
-                   :query-params)))))))
```
nwolfe, line 223: Moved the query-params test up because they were accidentally under the ACL testing block

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
camlow325: We'll need to reconcile this with the changes coming in #10, obviously.

src/puppetlabs/trapperkeeper/authorization/rules.clj
camlow325: The "golden rule".  :+1:

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -95,6 +81,18 @@
...
+      (throw (IllegalArgumentException.
+              (str "The authorization rule specified as " (pprint-rule rule)
+                   " does not contain a '" (name k) "' key.")))))
+  (when (or (not (integer? (:sort-order rule)))
+            (< (:sort-order rule) 1)
+            (> (:sort-order rule) 999))
+    (throw (IllegalArgumentException.
+            (str "The sort-order set in the authorization rule specified as "
+                 (pprint-rule rule) " is invalid. It should be a number "
+                 "between 1-999."))))
```
camlow325: Maybe say "from 1 to 999" since "between" might imply that 1 and 999 are invalid?

src/puppetlabs/trapperkeeper/authorization/rules.clj
```diff
@@ -154,6 +141,12 @@
            (str " at " file ":" (:line rule)))
          (format " (authentic: %s)" authentic?))))
 
+(schema/defn sort-rules :- Rules
+  "Sorts the rules based on their :sort-order, and then their :name if they
+   have the same sort order value."
+  [rules :- Rules]
+  (vec (sort-by (juxt :sort-order :name) rules)))
```
camlow325: (coolcoolcool)

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
nwolfe: Good call

src/puppetlabs/trapperkeeper/authorization/rules.clj
```diff
@@ -170,7 +163,8 @@
   [rules :- Rules
    request :- ring/Request
    name :- schema/Str]
-  (if-let [ { matched-rule :rule matches :matches } (some #(match? % request) rules)]
+  (if-let [{matched-rule :rule matches :matches}
+           (some #(match? % request) (sort-rules rules))]
```
camlow325: This might be a bit of a pre-optimization, but I'm wondering if it would be better to do the rule sort one time - behind `transform-code`, before they are stored into the Trapperkeeper service context - rather than here.  I think doing it here would mean that the rules would end up being resorted every time an authorization check is done?

camlow325: Looks like they do get sorted in `transform-code` anyway - but the pre-sorted value is what's stored in the TK service context rather than the value after sorting.  Maybe could just store the value after sorting and not have to do it again here?

nwolfe: Yeah, we could do that.  I did the sorting here to ensure that rules would be checked in sorted order, in the case where something somehow was added or done to the rules after `transform-code` (and we wouldn't know if they were sorted correctly anymore).

camlow325: I'm not seeing a need to modify the rules after the initial sort in `transform-code` is done.  Even if we add the ability to do a runtime reload of rules from disk later on, I think we'll be able to do that in such a way that the new rules can be sorted one time before replacing what is currently in the service context.  So, I'd like to pursue this change unless you see any major problems with it.

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -176,10 +174,21 @@
...
+  "Transforms the (validated) authorization service config into a list of Rules
+   that work with the authorization code. Assumes config has been validated via
+   `validate-auth-config!`. A warning is logged if the rules in the config are
+   not in ascending sort order."
   [config]
-  (map config->rule config))
+  (let [parsed (map config->rule config)
+        trim-fn #(select-keys % [:sort-order :name])]
+    (when-not (= (map trim-fn config) (map trim-fn (rules/sort-rules parsed)))
+      (log/warn "Rules in configuration file not in ascending sort order."))
```
camlow325: To help a user troubleshoot this further, would be nice to have the warning message indicate the first instance of a rule that is out of order - by rule name - depending upon how hard that would be to do.

camlow325:
:+1: Did some functional testing of this with Puppet Server and all seemed to be working as I'd expected.  Just a few comments for follow-up:
- Will need to consider conflicts around the `method` function with #10 if it comes in first.
- Minor wording update on the `sort-order` validation warning.
- Would like to see if we could do the rule sorting one time before the rules are stored in the TK context rather than each time a request is authorized.
- Would be nice to add at least the name of one offending rule when the rules from the configuration file are found to not be in ascending order.

src/puppetlabs/trapperkeeper/authorization/rules.clj
nwolfe: Sorting the code "over here" only so it has an effect "over there" feels smelly too though.  I don't see an immediate need to modify the rules after transform-code either, but that might change at some point and we'd have to remember the temporal coupling here.

I'll make the change, but I don't think we should be so quick to assume performance implications.

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
nwolfe: (challengeaccepted)

src/puppetlabs/trapperkeeper/authorization/rules.clj
dankreek: I'm not sure if `vec` is necessary here.

nwolfe: Hehe :smile: Yeah I think you're right

nwolfe: Oh right, I made it a vector because of the input+output schema `[Rule]`.

But, as I discovered recently, the schema `[Foo]` doesn't require the collection to be a vector, just something sequential.  So stripping off `vec` here satisfies the schema and everything works fine.  I was surprised when I discovered schema works this way, but I think it makes sense.

nwolfe:
@camlow325 Feedback addressed.

src/puppetlabs/trapperkeeper/authorization/rules.clj
camlow325: Thanks for doing that.  I think of the sorting as being one of the aspects of "normalization" that happens anytime we need to load/convert data from its external to its internal representation.  Ignoring the performance implications, it would seem artificial to me to only do "some" of that normalization at load time vs. other parts at execution time.  I'd have a really hard time envisioning that we'd ever want to do a resort after initial rule load/reload.  So, ultimately, I think this is the best way to go.

camlow325: One other thought I'd had on this is that we might consider having the initial construct that the rules are loaded into be a Clojure `sorted-set` rather than just a list/set and have the sorting be done implicitly as rules are added to the structure - similar to what we're already doing with the ACE (access control entries) that comprise the sorted ACL (access control list).  Could even do the validation for duplicates as part of the comparator so that we don't end up having to walk the list multiple times and question later on whether the validation has been completely done or not by the time we get to an arbitrary point in execution.  If we did this, we could have schema validate that the structure type implements `Sorted`, so we'd know that the sorting had to have been done on the input argument.  Not sure if the extra turmoil to go with something like this would be justified, though, so I'm not suggesting we necessarily add that to this PR.

camlow325:
Will need a rebase / resolve merge conflicts now that #10 has been merged.  After that, @nwolfe, I'm good with merging this assuming that @dankreek had had a chance to review this and hadn't brought anything else up.

src/puppetlabs/trapperkeeper/authorization/rules.clj
nwolfe: Hm, that sounds like conflating too many things, and I don't want schema to be responsible for any of it.  I'm not convinced that's better, but feel free to file a JIRA and we can come back to it.

camlow325: ok.  I won't bother filing a JIRA ticket for this.

nwolfe:
Feedback addressed.

@camlow325 Updated commit to only print the first mismatched rule rather than all of them.

camlow325:
:+1:

PR #13, opened 2015-09-15
(MAINT) Query params as keywords

nwolfe:
This commit fixes the mis-typing of query parameters keys from strings
to keywords. HOCON will parse the auth.conf file and return keywords,
not strings, for these query parameter fields.

### DIFF
--- a/src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
@@ -179,7 +179,7 @@
       (throw (IllegalArgumentException. "Rule query-params must be a map.")))
     (doseq [param (keys query-params)
             :let [value (get query-params param)]]
-      (when-not (string? param)
+      (when-not (keyword? param)
         (throw (IllegalArgumentException.
                 (str "The query-param '" param "' in the rule specified as "
                      (pprint-rule rule) " is invalid. It should be a string."))))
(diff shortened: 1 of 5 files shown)


camlow325:
:+1:

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -179,10 +179,10 @@
       (throw (IllegalArgumentException. "Rule query-params must be a map.")))
     (doseq [param (keys query-params)
             :let [value (get query-params param)]]
-      (when-not (string? param)
+      (when-not (keyword? param)
         (throw (IllegalArgumentException.
                 (str "The query-param '" param "' in the rule specified as "
-                     (pprint-rule rule) " is invalid. It should be a string."))))
+                     (pprint-rule rule) " is invalid. It should be a keyword."))))
```
nwolfe: This is kind of weird because they shouldn't actually be keywords in the file, they should be strings....

nwolfe: @camlow325 Wondering if I should delete this test...or change the invalid entry to be something other than a keyword or string, just so it doesn't look backwards and wrong

nwolfe: Whoops, meant to comment on the corresponding test for this

nwolfe:
@camlow325 Reverted "...should be a keyword" back to "...should be a string" and updated test

camlow325:
@nwolfe Reverting the error message back to "... should be a string" and associated test updates look good to me.

KevinCorcoran:
:see_no_evil:

PR #14, opened 2015-09-15
(TK-262) Add version # to auth.conf

nwolfe:
This commit adds a version number field to the top-level authorization
configuration map. The version number is only used during config
validation, after that it's discarded and the rule list is kept around
in the service context (like before).

Once we introduce a version 2, we'll have to persist the version number
through validation and condition some behavior on it.


test/puppetlabs/trapperkeeper/services/authorization/authorization_service_test.clj
```diff
@@ -44,7 +45,7 @@
 
   An empty list representing no rules with an expected behavior of
   default-deny originating in the authorization library itself."
-  {:authorization {:rules []}})
+  {:authorization {:version 1 :rules []}})
```
nwolfe, line 48: All other changes in this file are whitespace

test/puppetlabs/trapperkeeper/services/authorization/authorization_core_test.clj
```diff
@@ -69,6 +69,25 @@
...
+         IllegalArgumentException
+         #"Unsupported or missing version in configuration file.*"
+         (validate-auth-config! {:rules []}))))
+
+  (testing "Rule names must be unique"
+    (is (thrown-with-msg?
+         IllegalArgumentException
+         #".* Rules must be uniquely named."
+         (validate-auth-config! {:version 1
+                                 :rules [testrule testrule]}))))
```
nwolfe, line 92: I started to add a shared helper function in `testutils` that contained the top-level configuration map which could be shared by the service and core tests, but the result wasn't worth it.  If we add more tests and more defaults we could make the refactoring.

src/puppetlabs/trapperkeeper/services/authorization/authorization_service.clj
```diff
@@ -13,9 +13,9 @@
   [[:ConfigService get-in-config]]
   (init
     [this context]
-    (let [config (get-in-config [:authorization :rules])]
+    (let [config (get-in-config [:authorization])]
       (validate-auth-config! config)
-      (assoc-in context [:rules] (transform-config config))))
+      (assoc-in context [:rules] (transform-config (:rules config)))))
```
nwolfe, line 18: I chose not to change what we persist in the service's context since we only need the version number during validation.  We'll need to change this when we add code that needs the version number after validation.

dev-resources/puppetlabs/trapperkeeper/services/authorization/auth.conf
```diff
@@ -1,15 +0,0 @@
...
-   allow: [ "*.domain.org", "*.test.com" ]
-   deny: "bad.guy.com"
- },
- {
-   type: regex
-   path: "(incoming|outgoing)"
-   allow: "www.domain.org"
- }
- ]
-}
```
nwolfe, line 15: These files were never used.  We can add test fixtures in when we write the tests that use them.

test/puppetlabs/trapperkeeper/services/authorization/authorization_core_test.clj
```diff
@@ -69,6 +69,25 @@
...
+    (is (thrown-with-msg?
+         IllegalArgumentException
+         #"Missing authorization service configuration."
+         (validate-auth-config! nil))))
+
+  (testing "Version number required in configuration"
+    (is (thrown-with-msg?
+         IllegalArgumentException
+         #"Unsupported or missing version in configuration file.*"
+         (validate-auth-config! {:rules []}))))
```
nwolfe: I'm going to add two more iterations of this test against a non-zero number and a non-number value

src/puppetlabs/trapperkeeper/services/authorization/authorization_service.clj
camlow325, line 18: Works for me.

dev-resources/puppetlabs/trapperkeeper/services/authorization/auth.conf
camlow325, line 15: Sounds fine to me - wasn't aware we had these anyway.

camlow325:
:+1: @cprice404 - WDYT?

cprice404:
+1 but srsly on the whitespace thing

PR #15, opened 2015-09-16
(TK-280) Update README.md and add settings doc

camlow325:
This commit updates the `README.md` with some more specific information
about using tk-authz from the Trapperkeeper service perspective.  This
commit also introduces a new settings doc, `authorization-config.md`,
which covers the available settings that can be configured for
authorization.  As the new settings doc subsumes some of what had been
in the `README.md` doc, some of the language was removed from the
`README.md` doc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,116 +1,104 @@
-# Trapperkeeper-authorization
+# Trapperkeeper Authorization Service
 
 [![Build Status](https://travis-ci.org/puppetlabs/trapperkeeper-authorization.svg?branch=master)](https://travis-ci.org/puppetlabs/trapperkeeper-authorization)
 
-This clojure project is an authorization service for PuppetLabs Trapperkeeper.
-It aims to port Puppet's `auth.conf` feature to clojure, along with a different
-way to express authorization rules.
-
-The core of this service is a configuration format to express authorization
-rules which govern the REST API by parsing authentication information out of
-the incoming request, then matching the request against the configured rules.
-
-# Credits
-
-The original work for this library, service, and the original REST authconfig
-work in Ruby Puppet were all contributed by [Brice
-Figureau](https://github.com/masterzen).  This project has been graciously
-transferred to Puppet Labs for further development and maintenance as it
-becomes a critical part of the [Puppet
-Server](https://github.com/puppetlabs/puppet-server) security model as
-authconfig became a critical part of Puppet's security model.
+This project provides an authorization service for use with the
+[trapperkeeper service framework](https://github.com/puppetlabs/trapperkeeper).
+It aims to port Puppet's
+[`auth.conf`](https://docs.puppetlabs.com/puppet/latest/reference/config_file_auth.html)
+feature to Clojure and the trapperkeeper framework, with a different way to 
+express authorization rules.
 
 ## Installation
 
-Add the following to your _leiningen_ `project.clj`:
+To use this service in your trapperkeeper application, simply add this project
+as a dependency in your leiningen project file:
 
 [![Clojars Project](http://clojars.org/puppetlabs/trapperkeeper-authorization/latest-version.svg)](http://clojars.org/puppetlabs/trapperkeeper-authorization)
 
+Then add the authorization service to your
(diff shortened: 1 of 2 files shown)


camlow325:
@ahpook - would welcome your feedback on this one if you have time.

/CC @erikPrime and @mhaskel.

README.md
```diff
@@ -1,116 +1,104 @@
...
+
+The authorization service provides an implementation of the
+ `:AuthorizationService` interface.
+
+The authorization service is configured via the
+[trapperkeeper configuration service](https://github.com/puppetlabs/trapperkeeper/wiki/Built-in-Configuration-Service);
+so, you can control the authorization logic by adding an `authorization` section
+to one of your Trapperkeeper configuration files, and setting various 
+properties therein.  For more info, see
+[Configuring the Authorization Service](doc/authorization-config.md).
```
nwolfe, line 33: I almost feel like we should swap the information in this README and this config document.  Aren't people more interested in the `auth.conf` file rather than how the TK service works?  This documentation is very valuable, but shouldn't we target this README at users rather than devs?

camlow325, line 33: I intentionally modeled this off of how the `trapperkeeper-webserver-jetty9` docs are split up.  I tend to think that the docs in this repo should be a little more dev-oriented and that puppet-server end-user documentation will end up looking a little different - and live in the puppet-server repo.

nwolfe, line 33: Ah yeah, sure that makes sense

doc/authorization-config.md
```diff
@@ -0,0 +1,385 @@
...
+                    method: get
+                }
+                allow: "*"
+                sort-order: 1
+                name: "my_path"
+            },
+            {
+                match-request: {
+                    path: "/my_other_path"
+                    type: path
```
nwolfe, line 27: Would it be worthwhile to include a regex example?

doc/authorization-config.md
```diff
@@ -0,0 +1,385 @@
...
+                }
+                allow-unauthenticated: true
+                sort-order: 2
+                name: "my_other_path"
```
nwolfe, line 31: Would it be worthwhile to have a name with spaces?  Just thinking we could have our examples represent more of the configuration possibilities

doc/authorization-config.md
```diff
@@ -0,0 +1,385 @@
...
+
+~~~~hocon
+method: [ get, post ]
+~~~~
+
+Allowed values for `method` include `get`, `post`, `put`, `delete`, and `head`.
+
+##### `query-params`
+
+Optional.  If present, the value is should be a map of key/value pairs which
```
nwolfe: s/value is should/value should

doc/authorization-config.md
```diff
@@ -0,0 +1,385 @@
...
+[TK-268](https://tickets.puppetlabs.com/browse/TK-268).
+
+The supported forms for a `deny` entry are the same as those for an `allow` 
+entry.  See documentation for the [`allow`] (#allow) setting for more 
+information.
+
+#### `allow-unauthenticated`
+
+One of `allow`, `deny`, or `allow-unauthenticated` is required to be present
+for a rule.  If `allow-unauthenticated` is set to `true` for a rule, neither
```
nwolfe: The value doesn't matter, just whether `allow-unauthenticated` is present or not

nwolfe: Nope, I take that back, this line is correct :smile:

nwolfe:
:+1:

doc/authorization-config.md
camlow325, line 27: Yeah, that makes sense.  I altered one of the rules to be a `regex` in eef1f8e.

doc/authorization-config.md
camlow325, line 31: Good idea.  I altered one of the rules to have a space in the `name` in eef1f8e.

doc/authorization-config.md
camlow325: Yep, thanks.  Fixed in eef1f8e.

camlow325:
I rebased the PR and added a reference from the `README.md` to the recently integrated example in 2d3d856.

PR #16, opened 2015-09-17
(MAINT) Add simple ring authorization example

camlow325:
This commit adds a simple ring-based example which has a dependency on
the authorization service.  This example would just be used for
development purposes.


examples/ring_app/README.md
```diff
@@ -0,0 +1,152 @@
+# Simple Ring App with Authorization Example
+
+This example demonstrates how to incorporate the trapperkeeper-authorization
+service into a simple Ring app.  This is based loosely upon the
+[ring_app example] (https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/tree/master/examples/ring_app)
+in the trapperkeeper-webservices-jetty9 project.  See that example for more
```
MSLilah: s/webservices/webserver

examples/ring_app/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+## Launching trapperkeeper and running the app
+
+To start up _trapperkeeper_ and launch the sample application, use the
+following _lein_ command while in the _trapperkeeper-authorization_ home
+directory:
+
+~~~~sh
+lein trampoline run --config ./examples/ring_app/ring-example.conf \
+                    --bootstrap-config ./examples/ring_app/bootstrap.cfg
```
nwolfe, line 24: Considering that we don't expect any of this change, and that we can do it in the dev profile, I think we might as well add a lein alias that does, right?

examples/ring_app/README.md
```diff
@@ -0,0 +1,152 @@
...
+global: {
+    logging-config: ./examples/ring_app/logback.xml
+}
+
+webserver: {
+    port: 8080
+    ssl-port: 8081
+    ssl-cert: "./dev-resources/config/jetty/ssl/certs/localhost.pem"
+    ssl-ca-cert: "./dev-resources/config/jetty/ssl/certs/ca.pem"
+    ssl-key: "./dev-resources/config/jetty/ssl/private_keys/localhost.pem"
```
nwolfe: What do you think about putting these SSL files under the examples/ring_app to keep it all self-contained under one directory?

nwolfe:
:+1: this is awesome! Just wondering about the alias and moving SSL files

examples/ring_app/README.md
camlow325: I'm open to that.  I was thinking if we had multiple examples sometime down the road that sharing the files under `./dev-resources` might make sense.  But not sure how likely that would be anyway, and I agree that would make the example more self-contained.  I'll go ahead and change this.

MSLilah:
:+1:

examples/ring_app/README.md
camlow325: Thanks, fixed in 684ec19.

examples/ring_app/README.md
camlow325, line 24: Sounds good.  Added one called `ring-example` in 684ec19.

examples/ring_app/README.md
camlow325: I moved these to `./examples/ring_app/ssl` in 684ec19.

nwolfe:
:confetti_ball:

PR #17, opened 2015-09-22
(TK-282) Add `allow-header-cert-info` support

camlow325:
This commit adds support for the `allow-header-cert-info` feature, for
controlling whether the request identity should be read from an SSL
certificate (true) or HTTP X-Client headers (false).

### DIFF
--- a/doc/authorization-config.md
+++ b/doc/authorization-config.md
@@ -45,15 +45,13 @@ should use.  The only supported value is "1".
 
 ### `allow-header-cert-info`
 
-*(TBD, [SERVER-763](https://tickets.puppetlabs.com/browse/SERVER-763))*
-
 Optional.  Controls how the authenticated user "name" is derived for a 
 request being authorized.  Default value for the setting is `false`.
 
 For a value of `false`, the authenticated "name" for the request is derived 
 from the Common Name (CN) attribute within an X.509 certificate's Subject 
 Distinguished Name (DN).  The `wrap-with-authorization-check` middleware 
-tries to get the request's X.509 certificate from the `:ssl-client-cert` key 
+tries to get the request's X.509 certificate from the `ssl-client-cert` key
 in the Ring request map.  If the certificate cannot be found, e.g., if the 
 request was made over plaintext or was made over SSL/TLS but no certificate 
 was provided by the client, or the CN is not present in the certificate, the
@@ -62,13 +60,38 @@ request is considered "unauthenticated".
 For a value of `true`, the authenticated "name" for the request is derived from
 evaluating the values set for the `X-Client-DN` and `X-Client-Verify` HTTP 
 headers in the request.  The value for an `X-Client-DN` HTTP header should be
-in the form of a Subject DN from an X.509 certificate.  If the 
-`X-Client-Verify` HTTP header for the request has a value of `SUCCESS`, the 
-authenticated "name" for the request is extracted from the CN attribute in 
-the `X-Client-DN` value, if available.  If the `X-Client-Verify` HTTP header 
-is not present or does not have a value of `SUCCESS` and/or the CN cannot be 
-extracted from the `X-Client-DN` value, the request is considered
-"unauthenticated".
+in the form of a Subject DN from an X.509 certificate (e.g., `CN=myname`).  A
+value of `SUCCESS` for the `X-Client-Verify` HTTP header indicates that the
+request was validated successfully.  If the `X-Client-Verify` HTTP header is not
+present or does not have a value of `SUCCESS` and/or the CN cannot be extracted
+from the `X-Client-DN` value, the request is considered "unauthenticated".
+
+The `X-Client-DN` will attempt to be parsed first as an DN per
+[RFC 2253](https://www.ietf.org/rfc/rfc2253.txt).  For example:
+ 
+~~~~
+O=tester\, inc., CN=tester.test.org
+~~~~
+
+If the DN is not found to conform to the RFC 2253 format, the `X-Client-DN`
+will be parsed per the OpenSSL
+[compat](https://www.openssl.org/docs/manmaster/apps/x509.html) DN format, where
+attribute key/value pairs are delimited by solidus, `/`, characters:
+
+~~~~
+/O=tester, inc./CN=tester.test.org
+~~~~
+
+If a CN value cannot be derived via either parsing approach, the handler returns
+an HTTP 400/Bad Request response.
+
+> **Note:** The "compat" OpenSSL DN format does not provide a way to escape
+special characters in the DN.  If a solidus character were intended to be
+part of the value of an attribute, an unintended CN value could be derived.
+For example, the CN extracted from a DN of `/CN=tester/ inc.` is interpreted as
+`tester` rather than as `tester/ inc.`.  The RFC 2253 format has a specified
+approach for escaping special characters and is, therefore, preferred for
+expressing DN values, where possible.
 
 An "unauthenticated" request can only be "allowed" when the first matching 
 rule has an `allow-unauthenticated` setting with a value of `true`.  If 
@@ -79,10 +102,6 @@ request, the request is "denied" - in which case the handler returns an HTTP
 settings for the first rule which matches the request.  See the documentation
 for the [`allow`](#allow) and [`deny`](#deny) settings for more information.
 
-Until the work is completed for SERVER-763, trapperkeeper-authorization will
-only try to derive the authenticated "name" for a request from the CN in the
-X.509 certificate.
-
 ### `rules`
 
 Required.  An array in which each of the elements is a map of settings 
@@ -92,13 +111,13 @@ pertaining to a rule.  Here is an example of an array with two rules:
 rules: [
         {
             match-request: {
-                path: "/my_path"
-                type: path
+                path: "^/my_path/([^/]+)$"
+                type: regex
                 method: get
             }
-            allow: "*"
+            allow: "$1"
             sort-order: 1
-            name: "my_path"
+            name: "user-specific my_path"
         },
         {
             match-request: {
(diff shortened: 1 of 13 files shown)


test/puppetlabs/trapperkeeper/authorization/ring_middleware_test.clj
```diff
@@ -15,46 +20,287 @@
...
-    (testing "Denied when deny all"
-      (let [app (build-ring-handler [(-> (testutils/new-rule :path "/")
-                                         (rules/deny "*"))])]
-        (doseq [path ["a" "/" "/hip/hop/" "/a/hippie/to/the/hippi-dee/beat"]]
-          (let [req (testutils/request path :get "127.0.0.1" testutils/test-domain-cert)
-                {status :status} (app req)]
-            (is (= status 403))))))))
+      (ring-middleware/wrap-authorization-check rules allow-header-cert-info)))
+
+(deftest wrap-authorization-check-for-allow-header-cert-info-false-tests
```
camlow325, line 163: Most of the changes in here are just spacing - renamed from `wrap-authorization-check-test` to `wrap-authorization-check-for-allow-header-cert-info-false-tests`.  `build-ring-handler` now also called with extra argument, the `allow-header-cert-info` value.

test/puppetlabs/trapperkeeper/services/authorization/authorization_service_test.clj
```diff
@@ -1,17 +1,18 @@
...
-   [puppetlabs.trapperkeeper.authorization.testutils :refer :all]
-   [puppetlabs.trapperkeeper.services :refer [defservice]]
-   [puppetlabs.trapperkeeper.services.authorization.authorization-service
-    :refer [authorization-service]]
-   [puppetlabs.trapperkeeper.testutils.bootstrap :refer [with-app-with-config]]
-   [puppetlabs.trapperkeeper.testutils.logging :refer [with-test-logging]]
-   [ring.mock.request :as mock]
-   [ring.util.response :refer [response]]
-   [schema.test :as schema-test])
+    [clojure.test :refer :all]
```
camlow325, line 3: Mostly indentation changes.  :hand: :hand: :gun:  Also added the `puppetlabs.ssl-utils.core` ns reference.

doc/authorization-config.md
```diff
@@ -62,13 +60,38 @@ request is considered "unauthenticated".
...
-is not present or does not have a value of `SUCCESS` and/or the CN cannot be 
-extracted from the `X-Client-DN` value, the request is considered
-"unauthenticated".
+in the form of a Subject DN from an X.509 certificate (e.g., `CN=myname`).  A
+value of `SUCCESS` for the `X-Client-Verify` HTTP header indicates that the
+request was validated successfully.  If the `X-Client-Verify` HTTP header is not
+present or does not have a value of `SUCCESS` and/or the CN cannot be extracted
+from the `X-Client-DN` value, the request is considered "unauthenticated".
+
+The `X-Client-DN` will attempt to be parsed first as an DN per
```
nwolfe: s/an/a

test/puppetlabs/trapperkeeper/services/authorization/authorization_service_test.clj
```diff
@@ -230,6 +233,52 @@
...
+        (is (= "test.domain.org" (:name authorization)))
+        (is (= "test.domain.org" (ssl-utils/get-cn-from-x509-certificate
+                                  (:certificate authorization)))))))
+  (testing "Bad authorized user info generates bad request error"
+    (let [app (build-ring-handler default-rules
+                                  (assoc-in minimal-config
+                                            [:authorization
+                                             :allow-header-cert-info]
+                                            true))]
+      (let [req (-> base-request
```
nwolfe: Back-to-back `let`s in these three testing blocks could be combined

test/puppetlabs/trapperkeeper/authorization/testutils.clj
```diff
@@ -26,3 +31,9 @@
    (new-rule type path :any))
   ([type path method]
    (rules/new-rule type path method 500 "test rule")))
+
+(defn url-encoded-cert
+  [x509-cert]
+  (let [cert-writer (StringWriter.)
+        _ (ssl-utils/cert->pem! x509-cert cert-writer)]
+    (ring-codec/url-encode cert-writer)))
```
nwolfe, line 39: Could rewrite this without the `let` and `_` by using `->>`:

``` clj
(defn url-encoded-cert
  [x509-cert]
  (->> (StringWriter.)
       (ssl-utils/cert->pem! x509-cert)
       (ring-codec/url-encode)))
```

src/puppetlabs/trapperkeeper/authorization/ring.clj
```diff
@@ -2,6 +2,10 @@
   (:require [schema.core :as schema])
   (:import (java.security.cert X509Certificate)))
 
+(def cert-key
+  "The nested key where the client certificate for the request is stored"
+  [:authorization :certificate])
+
 (def is-authentic-key
   "The nested key where authenticity information is stored."
   [:authorization :authentic?])
```
nwolfe: I don't see hardly any value in these `def`s since we're not consistently using this pattern; we just dig into the map directly in most places.

nwolfe:
Nicely done as always :+1:

src/puppetlabs/trapperkeeper/authorization/ring.clj
camlow325: Where these end up getting exposed as "public API" - in the sense that code outside of tk-authz uses these keys via the Ring request map - I had been thinking it would be useful to have these as centralized constants.  I was even planning on referencing these directly in Puppet Server code as it pulls the keys back out.  Also, these serve as a way to provide a comment about what they are in a single point in the code.

That said, if you and @dankreek both think we should get rid of these and reference hardcoded names from the different places in code that use them, I could do that.  @dankreek - what do you think?

test/puppetlabs/trapperkeeper/authorization/testutils.clj
camlow325, line 39: I don't think I can do that, because `ssl-utils/cert->pem!` just returns back nil, and not the `writer`.

test/puppetlabs/trapperkeeper/services/authorization/authorization_service_test.clj
camlow325: Agreed, thanks.  I'll clean that up.

test/puppetlabs/trapperkeeper/authorization/testutils.clj
nwolfe, line 39: Oh right, duh :)  This is probably why I see "side-effecting" functions return the input in addition to modifying it, so you can use the function in a pipeline like this

camlow325, line 39: Yeah, that'd probably be a good thing to do with those functions at some point, I agree.  Maybe the next time we're in ssl-utils and happen to think about it.  Don't think I'd bother for now.

nwolfe, line 39: Yeah, not worth any time right now

test/puppetlabs/trapperkeeper/services/authorization/authorization_service_test.clj
camlow325: Addressed this in 85b9bb8.

doc/authorization-config.md
camlow325: Addressed this in 85b9bb8.

src/puppetlabs/trapperkeeper/authorization/ring.clj
dankreek: I think I'm more in favor of using functions for these kinds of things, which seems more idiomatic in the Ring world

dankreek: i.e. functions that do extract info from a ring request like `get-header` or all the other functions here: https://ring-clojure.github.io/ring/ring.util.request.html

camlow325: Ok, I'll look into doing the get/set functions instead.  For cases where there are error messages referencing the name of the header, suppose I'll just define them explicitly at the point they are referenced rather than referring to a constant.

camlow325: n/m the "header" comment - these are just Ring keys.  duh.

dankreek:
:+1: This is great work! Just had that one lingering comment about the decorator functions and it looks mergeable to me.

src/puppetlabs/trapperkeeper/authorization/ring.clj
camlow325: Replaced the key name constants with getter/setter Ring request functions in d90123b.

src/puppetlabs/trapperkeeper/authorization/ring.clj
```diff
@@ -23,3 +15,36 @@
    schema/Keyword schema/Any})
 
 ;; Functions
+
+(schema/defn get-authorized-authentic? :- schema/Bool
```
dankreek: I don't really think the `get-` part of these functions is necessary but not enough to hold up the merge. You want to hit the button, @nwolfe ?

camlow325: You mean, could call them `authorized-authentic?` instead of `get-authorized-authentic?`?  I could do that.

dankreek: Yeah, it lines up with the rest of Ring's functions which extract things from the request.

camlow325: Fair enough.  I'll do that.

camlow325: Done in ca59fea.  Shift-F6++

PR #19, opened 2015-10-01
(MAINT) Minor improvements to the config examples in docs

camlow325:
This commit fills out the text around some of the examples in the
`authorization-config.md` to hopefully make it a little more obvious
where each element of the config is specified relative to other elements
in the config.


nwolfe:
:+1:

PR #22, opened 2015-10-08
[FOR REVIEW ONLY - DO NOT MERGE] (TK-261) Rule matching logging

masterzen:
This is a partial implementation of [TK-261](https://tickets.puppetlabs.com/browse/TK-261) which for the moment tackles the match logging by producing a `Report` that describes as a data-structure what happened internally during an authorization. This was ported from my old tk-authz and reworked/fixed during the Puppet Contributor Summit.

Basically upon calling `rules/allowed?`, the caller will also receive a `report/Report` containing a description of the incoming request, the list of rules that were not matching, matching and skipped. And for the one rule that matched, an ACL report is returned, describing which ACE triggered.

Note: this PR is not finished as it doesn't provide anything to do the logging, and doesn't exploit new tk-authz feature like rules identifier.

Let me know if this is going in the right direction or not.

### DIFF
--- a/src/puppetlabs/trapperkeeper/authorization/acl.clj
+++ b/src/puppetlabs/trapperkeeper/authorization/acl.clj
@@ -1,6 +1,7 @@
 (ns puppetlabs.trapperkeeper.authorization.acl
   (:require [schema.core :as schema]
             [clojure.string :as str]
+            [puppetlabs.trapperkeeper.authorization.report :as report]
             [inet.data.ip :as ip]))
 
 ;; Schemas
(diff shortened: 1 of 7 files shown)


camlow325:
@masterzen Very sorry for not getting back to you earlier on this one.

As mentioned earlier, we very much appreciate all of the effort that you put into this PR and definitely see some advantages in terms of the level of detail that could be extracted from this to help with troubleshooting rule matching issues.  As you'd seen from the discussion on [TK-261](https://https://tickets.puppetlabs.com/browse/TK-261) and #23, we had ended up going with a simpler solution with less structural turmoil, albeit a much less granular/capable solution, for the first pass of this.  This had satisfied @ahpook's initial requirements / expectations for initial ticket.  So for now, we're thinking we'll close this PR and just stick with what's shipping in current code.

We'll continue monitoring feedback that we get from users about their experiences troubleshooting issues with tk-authorization and Puppet Server in general.  Should that show a need for a more comprehensive logging solution for tk-auth rule matches, I think we'll want to look at this further to see how it might help meet those needs.

Again, thanks so much for all of your help in developing the tk-authorization feature!

PR #23, opened 2015-10-09
(TK-261) Rule-by-rule logging 

nwolfe:
This commit adds logging during service start-up that prints the parsed
rules, and logging during request matching that shows each rule as it's
being evaluated.

---

You can see what the log output looks like here: https://gist.github.com/nwolfe/e278f6d4383ffe5578aa
- Start-up logging of parsed rules: https://gist.github.com/nwolfe/e278f6d4383ffe5578aa#file-puppetserver-log-L2-L104
- Rule-by-rule logging during matching: https://gist.github.com/nwolfe/e278f6d4383ffe5578aa#file-puppetserver-log-L179-L192

### DIFF
--- a/src/puppetlabs/trapperkeeper/authorization/rules.clj
+++ b/src/puppetlabs/trapperkeeper/authorization/rules.clj
@@ -111,6 +111,22 @@
             (some (get rule-params k)
                   (flatten [(get request-params (name k))])))))
 
+(schema/defn sort-rules :- [Rule]
+  "Sorts the rules based on their :sort-order, and then their :name if they
+   have the same sort order value."
+  [rules :- [Rule]]
+  (sort-by (juxt :sort-order :name) rules))
+
+(schema/defn requestor :- schema/Str
+  "Returns a string that identifies the source of the request containing
+   at least the IP address and the hostname if available."
+  [request :- ring/Request]
+  (let [ip (:remote-addr request)
+        name (ring/authorized-name request)]
+    (if (empty? name)
+      (str ip)
+      (format "%s(%s)" name ip))))
+
 (schema/defn match? :- (schema/maybe RuleMatch)
   "Returns the rule if it matches the request URI, and also
    any capture groups of the Rule pattern if there are any."
@@ -119,30 +135,48 @@
   (if (and (method-match? (:request-method request) (:method rule))
            (query-params-match? (:query-params request) (:query-params rule)))
     (if-let [matches (re-find* (:path rule) (:uri request))]
-      {:rule rule :matches (into [] (rest matches))})))
+      {:rule rule :matches (into [] (rest matches))}
+      (log/tracef
+       "Request to '%s' from '%s' did not match rule '%s' - continuing matching"
+       (:uri request) (requestor request) (:name rule)))))
 
 (defn- request->description
-  [request name rule]
-  (let [ip (:remote-addr request)
+  [request rule]
+  (let [from (requestor request)
         path (:uri request)
         method (:request-method request)
         authentic? (true? (ring/authorized-authentic? request))]
-    (str "Forbidden request: " (if name (format "%s(%s)" name ip) ip)
-         " access to " path " (method " method ")"
+    (str "Forbidden request: " from " access to " path " (method " method ")"
          (if-let [file (:file rule)] (str " at " file ":" (:line rule)))
          (format " (authentic: %s) " authentic?)
          (format "denied by rule '%s'." (:name rule)))))
 
-(schema/defn sort-rules :- [Rule]
-  "Sorts the rules based on their :sort-order, and then their :name if they
-   have the same sort order value."
-  [rules :- [Rule]]
-  (sort-by (juxt :sort-order :name) rules))
+(schema/defn allow-request :- AuthorizationResult
+  "Logs debugging information about the request and rule at the TRACE level
+   and returns an authorized authorization result with the provided message."
+  [request :- ring/Request
+   rule :- Rule
+   message :- schema/Str]
+  (log/tracef
+   "Request to '%s' from '%s' handled by rule '%s' - request allowed"
+   (:uri request) (requestor request) (:name rule))
+  {:authorized true
+   :message message})
 
 (schema/defn deny-request :- AuthorizationResult
-  "Logs the reason message at ERROR level and
-   returns an unauthorized authorization result."
-  [reason :- schema/Str]
+  "Logs debugging information about the request and rule at the TRACE level
+   as well as the reason for denial at the ERROR level, and returns an
+   unauthorized authorization result with the provided reason message."
+  [request :- ring/Request
+   rule :- (schema/maybe Rule)
+   reason :- schema/Str]
+  (if rule
+    (log/tracef
+     "Request to '%s' from '%s' handled by rule '%s' - request denied"
+     (:uri request) (requestor request) (:name rule))
+    (log/tracef
+     "Request to '%s' from '%s' did not match any rules - request denied"
+     (:uri request) (requestor request)))
   (log/error reason)
   {:authorized false
    :message reason})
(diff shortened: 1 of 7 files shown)


src/puppetlabs/trapperkeeper/authorization/ring.clj
```diff
@@ -38,13 +38,13 @@
...
   [request :- Request]
-  (get-in request [:authorization :name] ""))
+  (get-in request [:authorization :name]))
 
 (schema/defn set-authorized-name :- Request
   "Set the authorized client name onto the request."
   [request :- Request
    name :- (schema/maybe schema/Str)]
-  (assoc-in request [:authorization :name] (str name)))
+  (assoc-in request [:authorization :name] name))
```
nwolfe: The changes in this file rippled out to several files below - now we expect `nil` instead of `""`.  It turned out to be cleaner to not auto-convert `nil`<->`""` in these functions since there are now more places where `:name` is considered, and those need to differentiate between a `nil` name and a `""` name.

src/puppetlabs/trapperkeeper/authorization/rules.clj
```diff
@@ -125,67 +121,90 @@
...
+  [rules :- [Rule]]
+  (sort-by (juxt :sort-order :name) rules))
+
+(schema/defn requestor :- schema/Str
+  "Returns a string that identifies the source of the request containing
+   at least the IP address and the hostname if available."
+  [request :- ring/Request]
+  (if-let [name (ring/authorized-name request)]
+    (format "%s(%s)" name (:remote-addr request))
+    (str (:remote-addr request))))
```
nwolfe: This function is where we need to differentiate between a `:name nil` and `:name ""`.  The `(format ...)` line was extracted from `request->description` and was incorrectly returning the `(1.2.3.4)` when it should've been returning `1.2.3.4` in the case of no request `:name`.

It was incorrectly doing this because `ring/set-authorization-name` was converting the input from `nil`->`""`.

Anyway, much ado about nothing, just explaining why I got rid of the `nil`<->`""` conversion

src/puppetlabs/trapperkeeper/authorization/rules.clj
```diff
@@ -125,67 +121,90 @@
...
     (if (true? (:allow-unauthenticated rule))
-      {:authorized true :message "allow-unauthenticated is true - allowed"}
-      (if (and (true? (ring/authorized-authentic? request)) ; authenticated?
-            (acl/allowed? (:acl rule) name (:remote-addr request) matches))
-        {:authorized true :message ""}
-        (deny-request (request->description request name rule))))
-    (deny-request "global deny all - no rules matched")))
+      (allow-request request rule "allow-unauthenticated is true - allowed")
+      (if (and (true? (ring/authorized-authentic? request))
+               (acl/allowed? (:acl rule) (or (ring/authorized-name request) "")
```
nwolfe: This is the last spot where the `""` and `nil` is relevant.  I looked into `acl/allowed?` and it does internally support a `nil` name, but the function schema doesn't (i.e. `String` and not `(schema/maybe String)`). 

We could loosen the schema to support nil, which actually might make sense given the docstring of `acl/allowed?`, but I didn't want to bite that off with all the other changes in here.

test/puppetlabs/trapperkeeper/services/authorization/authorization_service_test.clj
```diff
@@ -213,7 +213,7 @@
           req (assoc catalog-request-nocert :body "Hello World!")
           {:keys [status body]} (app req)]
       (is (= status 403))
-      (is (= body (str "Forbidden request: (127.0.0.1) "
+      (is (= body (str "Forbidden request: 127.0.0.1 "
```
nwolfe: The parens only showed up in `actual` because of the inadvertent `nil`<->`""` behavior that I removed.

nwolfe:
If/when this is merged I'll file a PR against puppet-server's docs explaining how to activate the DEBUG/TRACE logging implemented here.  By default, these log messages won't show up unless the `logback.xml` is modified.

nwolfe:
Rebased

src/puppetlabs/trapperkeeper/authorization/ring.clj
camlow325: This is probably okay.  This value is consumed by Puppet Server upstream but it appears that it was tolerating a `nil` properly before anyway.  Generally speaking, I like to have my APIs that externally expose a String value to return an empty string instead of a nil - just to reduce the likelihood that the client might blow up if it isn't properly checking for nil before dereferencing.  If it were straightforward to keep this as an empty string, that would be my preference.  But I'm okay going with nil here.  @dankreek - WDYT?

camlow325: Actually my comment above would be more pertinent to the getter than the setter here.

src/puppetlabs/trapperkeeper/authorization/rules.clj
```diff
@@ -23,18 +23,14 @@
...
 
-(def Rules
-  "A list of rules"
-  [Rule])
-
 (def RuleMatch
-  "A match? result"
-  (schema/maybe {:rule Rule :matches [schema/Str]}))
+  "A `match?' result containing the matched rule and any regex capture groups."
+  {:rule Rule :matches [schema/Str]})
```
camlow325, line 28: Yeah, good call.

src/puppetlabs/trapperkeeper/authorization/rules.clj
camlow325: If it was all just for this, I'd probably instead have done something like:

``` clj
(let [name (ring/authorized-name request)]
  (if (empty? name)
     (str (:remote-addr request))
     (format "%s(%s)" name (:remote-addr request))))
```

Not great having the extra level of nesting but I do think it is valuable to avoid passing around `nil` when an empty string would generally work fine.

Again, I'm not too concerned about this, though.  Would like to get @dankreek's opinion.

src/puppetlabs/trapperkeeper/authorization/rules.clj
camlow325: If we leave it as an empty string and not potentially nil, nothing has to change here, right?  Ok, I'll shut up now...

src/puppetlabs/trapperkeeper/authorization/ring.clj
nwolfe: I didn't realize this was an API change necessarily, but you're right.  It would be easy to switch this back to an empty string now - I'll try that.

nwolfe:
Reverted the `nil`<->`""` changes

camlow325:
I'd be :+1: on bringing this change in.  We'll also want to take a hard look at #22 as it is also related to this ticket and offers some significant benefits beyond what we'd get with this PR - especially around the level of granularity captured in the comparison operations, which could be helpful for an even deeper level of troubleshooting.  If we decide to proceed with #22 and this PR happens to be integrated first, we may ultimately end up subverting some of the logging logic that this PR would bring in.

masterzen:
Feel free to drop #22: it might be more fine-grained than @nwolfe solution here but it is also much more complex and code is not as readable as I'd like it to be (and not yet finished, since there's only the report generation, not the logging).

src/puppetlabs/trapperkeeper/authorization/rules.clj
```diff
@@ -115,67 +111,91 @@
             (some (get rule-params k)
                   (flatten [(get request-params (name k))])))))
 
-(schema/defn match? :- RuleMatch
+(schema/defn sort-rules :- [Rule]
+  "Sorts the rules based on their :sort-order, and then their :name if they
+   have the same sort order value."
+  [rules :- [Rule]]
+  (sort-by (juxt :sort-order :name) rules))
```
dankreek, line 118: `(juxt)`++

dankreek:
:+1: I dig this. It's good.

nwolfe:
> Feel free to drop #22: it might be more fine-grained than @nwolfe solution here but it is also much more complex and code is not as readable as I'd like it to be (and not yet finished, since there's only the report generation, not the logging).

@masterzen Thanks again for your efforts and PR!  In the interest of time we're going to go with this for now, but we'll be taking a good look at what you've done in your PR regarding the information/granularity and potentially incorporate that into what we've got.

We're also interested in exploring the pattern you've suggested about building up a "report" data structure and then logging it all at once in a single spot rather than littering the various helper function with log statements.

masterzen:
@nwolfe, that plans sounds good to me :)

PR #24, opened 2015-10-09
(PE-7742) Add authorization-check function

rlinehan:
Previously, it was only possible to execute an authorization check using
`wrap-with-authorization-check`, which only works on Ring requests. This
commit factors out the authorization-checking logic into its own service
function, `authorization-check`.

With this function, it should be possible to perform an authorization check on
a servlet request.

### DIFF
--- a/src/puppetlabs/trapperkeeper/authorization/ring_middleware.clj
+++ b/src/puppetlabs/trapperkeeper/authorization/ring_middleware.clj
@@ -15,6 +15,12 @@
            (java.io StringReader)))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Schemas
+
+(def AuthResultWithRequest
+  (assoc rules/AuthorizationResult :req ring/Request))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Private
 
 (def header-cert-name
@@ -242,15 +248,25 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Public
 
+(schema/defn authorization-check :- AuthResultWithRequest
+  "A function that checks that the request is allowed by the provided rules.
+  Returns a map with the request with auth info added, whether the request is
+  authorized, and a message."
+  [request :- ring/Request
+   rules :- rules/Rules
+   allow-header-cert-info :- schema/Bool]
+  (let [req (add-authinfo request allow-header-cert-info)
+        name (ring/authorized-name req)]
+     (assoc (rules/allowed? rules req name) :req req)))
+
 (schema/defn wrap-authorization-check :- IFn
   "A ring middleware that checks the request is allowed by the provided rules"
   [handler :- IFn
    rules :- rules/Rules
    allow-header-cert-info :- schema/Bool]
   (fn [request]
-    (let [req (add-authinfo request allow-header-cert-info)
-          name (ring/authorized-name req)
-          {:keys [authorized message]} (rules/allowed? rules req name)]
+    (let [{:keys [authorized message req]}
+          (authorization-check request rules allow-header-cert-info)]
       (if (true? authorized)
         (handler req)
         (-> (ring-response/response message)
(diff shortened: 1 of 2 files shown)


src/puppetlabs/trapperkeeper/authorization/ring_middleware.clj
```diff
@@ -242,15 +248,25 @@
...
+(schema/defn authorization-check :- AuthResultWithRequest
+  "A function that checks that the request is allowed by the provided rules.
+  Returns a map with the request with auth info added, whether the request is
+  authorized, and a message."
+  [request :- ring/Request
+   rules :- rules/Rules
+   allow-header-cert-info :- schema/Bool]
+  (let [req (add-authinfo request allow-header-cert-info)
+        name (ring/authorized-name req)]
+     (assoc (rules/allowed? rules req name) :req req)))
```
rlinehan, line 260: Currently my implementation of getting this running with a servlet doesn't use the value of `:req` at all. I'm not sure whether it's really necessary to have it in what this function returns, but it could potentially be useful and it makes it easier for `wrap-authorization-check` to just use this function.

rlinehan:
I guess this could probably use some tests for `authorization-function`? I was a bit short on time and so put this up without any tests because I wanted to get it reviewed, but I can come back and add tests if y'all agree that's a good idea.

src/puppetlabs/trapperkeeper/authorization/ring_middleware.clj
camlow325, line 260: Sounds reasonable to me.

src/puppetlabs/trapperkeeper/services/authorization/authorization_service.clj
```diff
@@ -20,6 +21,9 @@
           (assoc-in [:allow-header-cert-info] (get config
                                                    :allow-header-cert-info
                                                    false)))))
+  (authorization-check [this request]
+   (let [{:keys [rules allow-header-cert-info]} (service-context this)]
+    (ring-middleware/authorization-check request rules allow-header-cert-info)))
```
camlow325, line 26: I think it would be good for calls to this function to have query-params destructured into the request - like is done for calls to `wrap-with-authorization-check` via the `wrap-query-params` middleware - before the request is authorized.  This would be useful in the event that the request needs to be validated against rules that might have `query-params` constraints on them.

Now that we'd have a couple of paths into the core middleware functions, I wonder if it would be better to just call `assoc-query-params` on the request - like the `ring-middleware/wrap-query-params` currently does - from inside of the new `ring-middleware/authorization-check` function.  This could be done to the request right before it is threaded through to the `rules/allowed?` call - and retained on the request that is returned in the response map from `ring-middleware/authorization-check`.

@rlinehan, @nwolfe, @dankreek - WDYT?

camlow325:
I'm pretty much :+1: on the changes - beyond the one comment I'd made about wrapping query params onto the request behind calls made into the new `authorization-check` function.  I do also think it would be good to add some documentation to the `README.md` for the new protocol function as well as a few tests for the new protocol function.

@cprice404 - I understand this may be time-sensitive from your team's perspective.  I'd also like to get this in soon as we're trying to wrap up our work toward a first release of tk-authz.  If @rlinehan is going to be out for a few days, would you like one of us to pick this up from what @rlinehan has done and finish it off?

src/puppetlabs/trapperkeeper/authorization/ring_middleware.clj
nwolfe, line 260: If we think the `req` might be useful I think we should make it a part of the existing `AuthorizationResult` rather than create a special schema like this.

That makes sense to me, for the `AuthorizationResult` to include information like the request that was authorized, and maybe later the rule that authorized it.

camlow325, line 260: @nwolfe You thinking, then, that `rules/allowed?` would add the `:req` into the results map that it returns?  Sounds reasonable to me.  Good for future expansion, too, like you mentioned.

nwolfe, line 260: Yeah, `rules/allowed?` would probably be the best place.  I was also thinking maybe the `rules/authorized?` function, but since `allowed?` is really the only function that does anything it's probably better there.

src/puppetlabs/trapperkeeper/services/authorization/authorization_service.clj
nwolfe, line 26: Yes, the core functions should not have to adjust or modify the request in any way.  And I agree we need to account for the query-params with this new function.  :+1: for having something in the HTTP/ring level do this before calling into core.

nwolfe:
Just FYI...As discussed I'm picking this up and will put up a new PR with Ruth's commit and a follow-up commit with the changes discussed here.

Closing this PR.

nwolfe:
Continued in #25

PR #25, opened 2015-10-13
(PE-7742) Add authorization-check function

nwolfe:
Previously, it was only possible to execute an authorization check using
`wrap-with-authorization-check`, which only works on Ring requests. This
commit factors out the authorization-checking logic into its own service
function, `authorization-check`.

With this function, it should be possible to perform an authorization check on
a servlet request.

---

This commit merges the `AuthResultWithRequest` into the existing
`AuthorizationResult`, so now all authorization results contain the
request in question.

Query parameter destructuring has now been moved into the new
`authorization-check` function, and the existing
`wrap-with-authorization-check` has been refactored to call it. This
allows for `wrap-with-query-params` to be removed.

Added a test for the new `authorization-check` function which also
covers query parameter support and the newly returned (and updated)
request in the authorization result.

### DIFF
--- a/src/puppetlabs/trapperkeeper/authorization/ring_middleware.clj
+++ b/src/puppetlabs/trapperkeeper/authorization/ring_middleware.clj
@@ -15,12 +15,6 @@
            (java.io StringReader)))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
-;;; Schemas
-
-(def AuthResultWithRequest
-  (assoc rules/AuthorizationResult :req ring/Request))
-
-;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Private
 
 (def header-cert-name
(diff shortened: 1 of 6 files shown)


nwolfe:
Ping @camlow325 @rlinehan for review

src/puppetlabs/trapperkeeper/authorization/ring_middleware.clj
```diff
@@ -202,8 +202,8 @@
 
 (schema/defn add-authinfo :- ring/Request
   "Add authentication information to the ring request."
-  [request :- ring/Request
-   allow-header-cert-info :- schema/Bool]
+  [allow-header-cert-info :- schema/Bool
+   request :- ring/Request]
```
nwolfe, line 206: I swapped these parameters just for the convenience of being able to thread the request in a `->>` pipeline in `authorization-check`

nwolfe:
CC @cprice404 @fpringvaldsen just because ye might have been waiting on these changes

cprice404:
:see_no_evil: @fpringvaldsen can you test your branch of file sync against this and confirm it still works for you?  It doesn't look like @nwolfe 's commit changes the public API.

@nwolfe just FYI @rlinehan is out this week, so we'll just get it reviewed / merged between myself, @fpringvaldsen , and @camlow325 .

camlow325:
:+1: looks good to me.

Only thing I'd suggest is that it would be good to have at least one test added to the `authorization-service-test` namespace which exercise the new `authorization-check` function at the Trapperkeeper service function.  I don't think this would need to duplicate all of what the lower-level tests are doing for this in the `ring-middleware` namespace.  I'd mostly just be interested in seeing that calls to the service function are getting through to the function in the `ring-middleware` namespace and are returning reasonable looking results.  To do this without mocking/redefing, maybe we could just have one authorized and one non-authorized test case to confirm that the plumbing is functional.

MSLilah:
@cprice404 I tested this with my branch, and all tests are passing

nwolfe:
@camlow325 Service API test added.

Ready for review / merge

camlow325:
:+1: - can wait a bit to see if it passes Travis before merging.

PR #27, opened 2015-10-29
(TK-286) Rename :authentic? to :authenticated

dankreek:


src/puppetlabs/trapperkeeper/authorization/rules.clj
```diff
@@ -149,10 +149,10 @@
   (let [from (requestor request)
         path (:uri request)
         method (:request-method request)
-        authentic? (true? (ring/authorized-authentic? request))]
+        authenticated? (true? (ring/authorized-authentic? request))]
```
camlow325: I was thinking we might as well change the `authorized-authentic?` ring helpers to `authorized-authenticated`.  Sound okay?

camlow325:
:+1: LGTM.  Only other change I can see is that we should also update the `README.md` file to refer to `authenticated` instead of `authentic?` where appropriate.

dankreek:
Docs are updated

camlow325:
:+1: @nwolfe - you good with this?

src/puppetlabs/trapperkeeper/authorization/ring.clj
```diff
@@ -16,16 +16,16 @@
...
 ;; Functions
 
-(schema/defn authorized-authentic? :- schema/Bool
+(schema/defn authorized-authenticated :- schema/Bool
   "Get whether the authorized client is considered authentic or not."
   [request :- Request]
-  (get-in request [:authorization :authentic?]))
+  (get-in request [:authorization :authenticated]))
 
 (schema/defn set-authorized-authentic? :- Request
```
nwolfe: We should change this function name too

nwolfe: And since we're in here again might as well update the docstrings on these two functions with s/authentic/authenticated

dankreek: A'ight, this should be good

src/puppetlabs/trapperkeeper/authorization/ring.clj
```diff
@@ -16,16 +16,16 @@
...
-  "Get whether the authorized client is considered authentic or not."
+(schema/defn authorized-authenticated :- schema/Bool
+  "Get whether the authorized client is considered authenticated or not."
   [request :- Request]
-  (get-in request [:authorization :authentic?]))
+  (get-in request [:authorization :authenticated]))
 
-(schema/defn set-authorized-authentic? :- Request
-  "Set whether the authorized client is considered authentic or not."
+(schema/defn set-authorized-authenticated? :- Request
```
nwolfe: Since these functions are supposed to be symmetrical and we got rid of the `?` everywhere else, we should remove it from this function name as well.

:metal:

src/puppetlabs/trapperkeeper/authorization/ring_middleware.clj
```diff
@@ -208,7 +208,7 @@
     (->
       request
       (ring/set-authorized-name name)
-      (ring/set-authorized-authentic? (and
+      (ring/set-authorized-authenticated? (and
                                        (verified? request
```
nwolfe: And if we're going back in for changes I suppose we should fix up the whitespace here too

dankreek:
@nwolfe allright. I think it's about that time. Let's do this.

PR #28, opened 2015-11-03
(TK-285) Fix the back reference thing

dankreek:


src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -155,12 +155,24 @@
                  (str/join "', '" (sort valid-methods)) "'"))))
   (when (= "regex" (-> rule :match-request :type name str/lower-case))
     (try
-      (re-pattern (:path (:match-request rule)))
+      (let [pattern (re-pattern (:path (:match-request rule)))
+            group-count (-> pattern (.matcher "") .groupCount)
```
camlow325, line 159: Nice

dankreek, line 159: Admittedly, it's a hack I found on Stack Overflow, but it seems to get the job done.

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -155,12 +155,24 @@
...
+            no-brs? (fn [matches]
+                      (not-any? string? (map #(re-find #"\$\d" %)
+                                             (flatten [(or matches "")]))))
+            has-brs (not (or (no-brs? (:allow rule))
+                             (no-brs? (:deny rule))))]
+        (if (and (not has-brs) (= group-count 0))
```
camlow325: I was hoping we could use something here that would be more comprehensive.  This would appear to help for the case where an allow or deny directive has a backreference and there are no capture groups in the regex.  But the bad error message could also come up in the case where the regex does have capture groups but a backreference refers to a non-existent capture group.

For example, I might have "$1 blah $6 blah $2" in an allow ACE with a regex of "^/puppet/v3/catalog/([^/]+)/(.*)$".  This regex has two capture groups so the $1 and $2 backreferences are valid but the $6 one is not.  I was hoping we could display the friendly, validation-time error for this case as well.

I was thinking that the algorithm would walk through each of the ACEs in the allow and deny directives and flag an error for each one found which exceeds the number of available capture groups.  For simplicity, if it were to just error out after the first one found which exceeds the number of capture groups, I'd be fine with that, too, although showing all of the offending ACEs could be a little more helpful for a user.

Does this make sense?

test/puppetlabs/trapperkeeper/services/authorization/authorization_core_test.clj
```diff
@@ -214,7 +214,16 @@
          IllegalArgumentException
          #".* 'delete', 'get', 'head', 'post', 'put'"
          (validate-auth-config-rule!
-          (assoc-in testrule [:match-request :method] ["nasty" "gross"]))))))
+          (assoc-in testrule [:match-request :method] ["nasty" "gross"])))))
+
+    (is (thrown-with-msg?
```
camlow325, line 246: Seems like this one shouldn't fail with the "at least one capture group" message.  A regex path isn't required to have capture groups.  In this case, there are no capture groups and none of the ACEs make any backreferences.  So this one looks valid to me.

camlow325:
Done reviewing for now.  Will review again after updates.

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
dankreek: I certainly didn't intend to dash your hopes. I'll see what I can do here.

test/puppetlabs/trapperkeeper/services/authorization/authorization_core_test.clj
```diff
@@ -105,116 +105,153 @@
...
 
     (is (thrown-with-msg?
-         IllegalArgumentException
-         #".* 'delete', 'get', 'head', 'post', 'put'"
-         (validate-auth-config-rule!
-          (assoc-in testrule [:match-request :method] "gross"))))
+          IllegalArgumentException
+          #".* 'delete', 'get', 'head', 'post', 'put'"
+          (validate-auth-config-rule!
+            (assoc-in testrule [:match-request :method] "gross"))))
```
dankreek, line 211: I can go back and fix this if it will end the world... otherwise I'm ok with this if ya'll are.

nwolfe, line 211: :ok_hand:

camlow325, line 211: (aaaaahhhhhhhh)

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -155,12 +155,31 @@
                  (str/join "', '" (sort valid-methods)) "'"))))
   (when (= "regex" (-> rule :match-request :type name str/lower-case))
     (try
-      (re-pattern (:path (:match-request rule)))
+      (let [pattern (re-pattern (:path (:match-request rule)))
+            group-count (-> pattern (.matcher "") .groupCount)
+            get-br-list (fn [match-str]
+                          (map #(Integer/parseInt (nth % 1))
+                               (re-seq #"\$(\d)" match-str)))
```
camlow325: I think this is missing a `+` - `#"\$(\d+)"`  This would help catch the error for backreferences larger than 9.

camlow325:
:+1: other than the missing `+` on the regex.  All seemed to work as I expected from some functional testing with the ring-example.  @nwolfe - WDYT?

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
dankreek: Ohhhh yes, of course

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -155,12 +155,31 @@
...
+            largest-br (fn [match]
+                         (let [br-list (flatten (map get-br-list
+                                                     (flatten [(or match "")])))]
+                           (when (> (count br-list) 0)
+                             (apply max br-list))))
+            largest-allow-br (or (largest-br (:allow rule)) 0)
+            largest-deny-br (or (largest-br (:deny rule)) 0)]
+        (doseq [[largest field] [[largest-allow-br "allow"]
+                                 [largest-deny-br "deny"]]]
+          (if (> largest group-count)
```
dankreek, line 172: Just gonna' say ^^ That is a pug of an algorithm.

nwolfe, line 172: Would it be worthwhile to extract into it's own function then? With a docstring explaining what it's doing?

I mean, if the author himself is calling it a pug of an algorithm...are we going to know what the pug it does in 6 months?

nwolfe:
:+1: :metal:

PR #29, opened 2015-11-04
(TK-289) Remove rule out-of-order warning

nwolfe:
This warning is more likely to cause confusion than be useful. Rather
than dumping a paragraph of text to clarify this just removes it
altogether.


nwolfe:
@camlow325 @rlinehan

camlow325:
:+1:

PR #30, opened 2015-11-05
(MAINT) Extract validate-regex-backreferences!

nwolfe:
This commit extract the regex back-reference & capture group validation
code from validate-auth-config-rule! into its own function.

### DIFF
--- a/src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
@@ -92,6 +92,42 @@
       :else
       false)))
 
+(defn validate-regex-backreferences!
+  "Validates that allow and deny back-references actually exist as capture
+   groups in the path regex. Throws an exception if a back-reference refers
+   to a capture group that doesn't exist."
+  [rule]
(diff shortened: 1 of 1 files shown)


nwolfe:
@dankreek t'was bugging me not in it's own function

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -92,6 +92,42 @@
...
+   to a capture group that doesn't exist."
+  [rule]
+  (let [group-count (try
+                      (-> rule
+                          :match-request
+                          :path
+                          re-pattern
+                          (.matcher "")
+                          .groupCount)
+                      (catch PatternSyntaxException e
```
nwolfe, line 107: This was originally around the whole function body, but only these first two lines could throw this.
I moved the catch up around only those lines, and then combined the two lines into a `->`.

Happy to reformat this any way, mainly just wanted to reduce the scope of the try/catch

camlow325:
:+1: @dankreek - any concerns?

dankreek:
:+1: I can dig it!

PR #33, opened 2016-03-28
(maint) Code Clean Up

vilmibm:
This PR largely cleans up use of schema and removes some dead code.
It mostly deals with the Entry schema but also does some other cleanup.
- Remove  vestigial :length key from Entry
- Removes superfluous :qualifier key from Entry
- Redesigns the :type key as :match in Entry
- Don't require clojure.string as str
- Other minor stuff

I did this because tk-293's implementation was getting hairy so I went
back to master and laid this groundwork to build on.

### DIFF
--- a/src/puppetlabs/trapperkeeper/authorization/acl.clj
+++ b/src/puppetlabs/trapperkeeper/authorization/acl.clj
@@ -11,7 +11,6 @@
   {:auth-type auth-type
   :type (schema/enum :domain :opaque :regex :dynamic)
   :qualifier (schema/enum :exact :inexact)
-  :length (schema/maybe schema/Int)
   :pattern schema/Any})
 
 (def ACL
@@ -59,7 +58,6 @@
     {:auth-type type
      :type :regex
      :qualifier :exact
-     :length nil
      :pattern #"^*$"}
 
     ; exact domain
@@ -67,7 +65,6 @@
     {:auth-type type
      :type :domain
      :qualifier :exact
-     :length nil
      :pattern (munge-name pattern)}
 
     ; *.domain.com
@@ -76,7 +73,6 @@
       {:auth-type type
        :type :domain
        :qualifier :inexact
-       :length (count host_sans_star)
        :pattern host_sans_star})
 
     ; backreference
@@ -84,7 +80,6 @@
     {:auth-type type
      :type :dynamic
      :qualifier :exact
-     :length nil
      :pattern (munge-name pattern)}
 
     ; opaque string
@@ -92,7 +87,6 @@
     {:auth-type type
      :type :opaque
      :qualifier :exact
-     :length nil
      :pattern [pattern]}
 
     ; regex
@@ -100,7 +94,6 @@
     {:auth-type type
      :type :regex
      :qualifier :inexact
-     :length nil
      :pattern (str/replace pattern #"^/(.*)/$" "$1")}
 
     :else
--- a/src/puppetlabs/trapperkeeper/authorization/acl.clj
+++ b/src/puppetlabs/trapperkeeper/authorization/acl.clj
@@ -4,10 +4,10 @@
 ;; Schemas
 
 (def AuthType (schema/enum :allow :deny))
-(def EntryPattern (schema/pred #(some nil? [(schema/check String %)
-                                            (schema/check [String] %)
-                                            (schema/check schema/Regex %)])
-                               "Entry Pattern must be a string, list of strings, or Regex"))
+(def EntryPattern (schema/conditional
+                   string? String
+                   sequential? [String]
+                   :else schema/Regex))
 
 (def Entry
   "An authorization entry matching a network or a domain"
--- a/src/puppetlabs/trapperkeeper/authorization/acl.clj
+++ b/src/puppetlabs/trapperkeeper/authorization/acl.clj
@@ -48,14 +48,13 @@
 
 (schema/defn ^:private split-domain :- [String]
   "Given a domain string, split it on '.' and reverse it. For examples,
-  'sylvia.plath.net' becomes ['net' 'plath' 'sylvia']. This is used for domain
+  'sylvia.plath.net' becomes ('net' 'plath' 'sylvia'). This is used for domain
   matching."
-  [pattern :- String]
-  (-> pattern
+  [domain :- String]
+  (-> domain
       (clojure.string/lower-case)
       (clojure.string/split #"\.")
-      reverse
-      vec))
+      reverse))
 
 (schema/defn new-domain :- Entry
   "Creates a new ACE for a domain"
@@ -125,7 +124,7 @@
   (if (= (:match ace) :backreference)
     (new-domain (ace :auth-type)
                 (clojure.string/join "." (map #(substitute-backreference % captures)
-                                   (reverse (ace :pattern)))))
+                                              (reverse (ace :pattern)))))
     ace))
 
 (schema/defn match? :- schema/Bool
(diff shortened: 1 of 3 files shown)


KevinCorcoran:
> Don't require clojure.string as str

Why?

src/puppetlabs/trapperkeeper/authorization/acl.clj
```diff
@@ -1,17 +1,19 @@
 (ns puppetlabs.trapperkeeper.authorization.acl
-  (:require [schema.core :as schema]
-            [clojure.string :as str]))
+  (:require [schema.core :as schema]))
 
 ;; Schemas
 
-(def auth-type (schema/enum :allow :deny))
+(def AuthType (schema/enum :allow :deny))
+(def EntryPattern (schema/pred #(some nil? [(schema/check String %)
```
KevinCorcoran: Could this use `schema/conditional`?

src/puppetlabs/trapperkeeper/authorization/acl.clj
```diff
@@ -44,57 +46,58 @@
...
 ;; ACE creation
 
-(defn munge-name
-  [pattern]
-  (-> pattern (str/lower-case) (str/split #"\.") reverse vec))
+(schema/defn ^:private split-domain :- [String]
+  "Given a domain string, split it on '.' and reverse it. For examples,
+  'sylvia.plath.net' becomes ['net' 'plath' 'sylvia']. This is used for domain
+  matching."
+  [pattern :- String]
```
KevinCorcoran: Perhaps rename this ... I would expect a parameter named `pattern` to be a regex.  I _think_ that's a convention used elsewhere, but I could be wrong.

src/puppetlabs/trapperkeeper/authorization/acl.clj
```diff
@@ -44,57 +46,58 @@
...
+  (-> pattern
+      (clojure.string/lower-case)
+      (clojure.string/split #"\.")
+      reverse
+      vec))
```
KevinCorcoran: Why is `vec` needed here?

src/puppetlabs/trapperkeeper/authorization/acl.clj
```diff
@@ -103,29 +106,25 @@
...
 (defn interpolate-backreference
   "change all possible backreferences in ace patterns to values from the
   capture groups"
   [ace captures]
-  (if (= (ace :type) :dynamic)
+  (if (= (:match ace) :backreference)
     (new-domain (ace :auth-type)
-                (str/join "." (map #(substitute-backreference % captures)
+                (clojure.string/join "." (map #(substitute-backreference % captures)
                                    (reverse (ace :pattern)))))
```
KevinCorcoran: I think the indentation on this line is incorrect now.

vilmibm:
@KevinCorcoran re: str, I think it's bad form to introduce a ns binding that is the same as an in scope function. I know clojure is smart enough to prevent any shadowing but it still strikes me as an anti-pattern. if you think `clojure.string` is too verbose, i'm happy to switch those to `:refer`

KevinCorcoran:
Oh yeah, I didn't realize there was a collision with `str`.  Good point.

src/puppetlabs/trapperkeeper/authorization/acl.clj
vilmibm: I hate this name. I was putting off thinking of something new until I got back to TK-293 proper but was considering something vague like "value" :/

src/puppetlabs/trapperkeeper/authorization/acl.clj
vilmibm: This is old code, I just left it in. Perhaps the original author thought reverse was lazy, but it's not; I'll take it out.

src/puppetlabs/trapperkeeper/authorization/acl.clj
vilmibm: An argument in favor of `pattern` is that it's storing "the thing to which another thing has to correspond" which seems like a proper use of the word. However, in our discipline, I agree that it almost always means "regex" which it only means here sometimes. I'm into `cynosure` but surely that is a smartass idea.

jpinsonault: maybe just call it domain?

KevinCorcoran: yeah, `domain` or `domain-string` seems to correspond with the docstring

src/puppetlabs/trapperkeeper/authorization/acl.clj
```diff
@@ -4,10 +4,10 @@
...
 
 (def AuthType (schema/enum :allow :deny))
-(def EntryPattern (schema/pred #(some nil? [(schema/check String %)
-                                            (schema/check [String] %)
-                                            (schema/check schema/Regex %)])
-                               "Entry Pattern must be a string, list of strings, or Regex"))
+(def EntryPattern (schema/conditional
+                   string? String
+                   sequential? [String]
+                   :else schema/Regex))
```
KevinCorcoran, line 10: yay, much nicer

KevinCorcoran:
:+1:

src/puppetlabs/trapperkeeper/authorization/acl.clj
```diff
@@ -45,63 +46,57 @@
 
 ;; ACE creation
 
-(defn munge-name
-  [pattern]
-  (-> pattern (str/lower-case) (str/split #"\.") reverse vec))
+(schema/defn ^:private split-domain :- [String]
+  "Given a domain string, split it on '.' and reverse it. For examples,
+  'sylvia.plath.net' becomes ['net' 'plath' 'sylvia']. This is used for domain
```
haus: without the `vec` call, this docstring should probably be updated to `('net' 'plath' 'sylvia')`

vilmibm: good catch. thank u clojure for having so many sequential types.

MSLilah:
:see_no_evil: I don't feel I have enough context to comment on the redesign of the `:type` key, but everything else looks good to me.

haus:
ditto to what @fpringvaldsen said.

src/puppetlabs/trapperkeeper/authorization/acl.clj
```diff
@@ -1,18 +1,19 @@
...
+  (:require [schema.core :as schema]))
 
 ;; Schemas
 
-(def auth-type (schema/enum :allow :deny))
+(def AuthType (schema/enum :allow :deny))
+(def EntryPattern (schema/conditional
+                   string? String
+                   sequential? [String]
+                   :else schema/Regex))
```
camlow325, line 10: Is `sequential? [schema/Regex]` not possible?  From an end user config perspective, I think you can have a list of regexes in an allow or deny ACL but I can't remember now how that maps onto the existing code.

camlow325, line 10: Oh, this is the pattern for an individual `Entry`, right.  In that case, I think my comment doesn't apply.  Suppose the `sequentlal? [String]` is there because tk-auth parses a domain string ACE out into that for its run-time use.

src/puppetlabs/trapperkeeper/authorization/acl.clj
masterzen: I don't exactly remember the original purpose, but I think it performed way better as a vec than a sequence.

PR #34, opened 2016-04-05
(TK-293) Implement SSL extension authorization rules

vilmibm:
This PR adds support for a new feature: The ability to write
authorization rules based on SSL Extensions (also known as CSR
Attributes).

This entails:
- Updating test utils to generate certs with extensions signed in more easily
- Some code cleanup
- Updating the configuration syntax to take a data structure instead of
  just string certnames when setting allow/deny values
- Configurable translation from OID -> short name when extracting
  extensions from certificates
- Updating the authorization enforcement code to match on either
  certname or extensions
- General refactoring to support a map where we used to only support a
  string.
- Updating tests to handle the new match data structure
- A flurry of new schemas

This only implements basic string matching for SSL extensions. In other words, extensions that are more complex than key -> value are not currently supported. You can, however, specify a list of valid extension values to match against.

### DIFF
--- a/doc/authorization-config.md
+++ b/doc/authorization-config.md
@@ -443,6 +443,44 @@ authorization: {
 }
 ~~~~
 
+... or a single map value with an extension key ...
+
+~~~~hocon
+authorization: {
+    version: 1
+    rules: [
+            {
+                match-request: {...}
+                allow: {
+                    extensions: {
+                        my_ssl_extension: some_value
+                    }
+                }
+                sort-order: 1
+                name: "my path"
+            }
+    ]
+}
+~~~~
+
+... or a single map value with a certname key (equivalent to a bare string) ...
+
+~~~~hocon
+authorization: {
+    version: 1
+    rules: [
+            {
+                match-request: {...}
+                allow: {
+                    certname: node1
+                }
+                sort-order: 1
+                name: "my path"
+            }
+    ]
+}
+~~~~
+
 ... or as an array of values ...
 
 ~~~~hocon
@@ -451,7 +489,7 @@ authorization: {
     rules: [
             {
                 match-request: {...}
-                allow: [ node1, node2, node3 ]
+                allow: [ node1, node2, node3, {extensions:{ext_shortname1: value1, ext_shortname2: value2}} ]
                 sort-order: 1
                 name: "my path"
             }
@@ -494,6 +532,101 @@ One of the following forms may be used as the value for an `allow` entry:
  of `xyz.domain.org` and a request URL of
  `"http://my-host:8080/the/path/xyz"`, however, would not be considered a 
  match for the entry.
+ 
+ * A map with either a `certname` or `extensions` key. The former behaves
+   exactly as a standalone string. The latter is used to specify a set of X.509
+   extensions to match. Each key specified in the extensions map must appear in
+   the request and the values are matched exactly (no wildcarding). Extensions
+   in the request not explicitly listed here are ignored. Note that a given
+   extension map won't match if not **all** keys match. If you want an "or"
+   relationship between extensions, split them into different extension maps.
+   You may also use a list of values when writing an extension map. A request
+   will match as long as it has an extension value that appears in the list for
+   a given key.
+ 
+   For example, given the rules:
+ 
+   ~~~~hocon
+   allow: [{extensions: {role: "compilemaster"
+                         env: "prod1"}}
+           {extensions: {role: console"
+                         env: ["appgroup1", "prod1"]}}],
+   deny: [{extensions: {env: ["test", "dev"]}}
+          {extensions: {role: "puppetdb"}}
+          {extensions: {role: "console", pp_env: "demo"}}]
+   ~~~~
+ 
+   Requests with the following extension maps would be **denied**:
+
+   ~~~~clojure
+   ;; Denied by env: test rule
+   {:role "compilemaster"
+    :env "test"
+     ;; irrelevant keys elided
+   }
+   ~~~~
+ 
+   ~~~~clojure
+   ;; Denied since this combo of role and env does not appear in an allow rule
+   {:role "compilemaster"
+    :env "appgroup2"
+    ;; irrelevant keys elided
+   }
+   ~~~~
+
+   ~~~~clojure
+   ;; Denied by role: puppetdb rule
+   {:role "puppetdb"
+    :env "prod1"
+    ;; irrelevant keys elided
+   }
+   ~~~~
+
+   ~~~~clojure
+   ;; Denied since role nor env keys are never matched by an auth rule
+   {:role "mco"
+    :env "prod1"
+    ;; irrelevant keys elided
+   }
+   ~~~~
+
+   ~~~~clojure
+   ;; Denied since experimental is neither appgroup1 nor prod1
+   {:role "console"
+    :env "experimental"
+    ;; irrelevant keys elided
+   }
+   ~~~~
+
+   The following request extension maps would be **allowed**:
+
+   ~~~~clojure
+   {:role "compilemaster"
+    :env "prod1"
+    ;; irrelevant keys elided
+   }
+   ~~~~
+
+   ~~~~clojure
+   {:role "console"
+    :env "prod1"
+    ;; irrelevant keys elided
+   }
+   ~~~~
+
+   ~~~~clojure
+   {:role "console"
+    :env "appgroup1"
+    ;; irrelevant keys elided
+   }
+   ~~~~
+
+   ~~~~clojure
+   {:role "console"
+    :env "appgroup1"
+    ;; irrelevant keys elided
+   }
+   ~~~~
 
 #### `deny`
 
@@ -518,6 +651,44 @@ authorization: {
 }
 ~~~~
 
+... or a single map value with an extension key ...
+
+~~~~hocon
+authorization: {
+    version: 1
+    rules: [
+            {
+                match-request: {...}
+                deny: {
+                    extensions: {
+                        my_ssl_extension: some_value
+                    }
+                }
+                sort-order: 1
+                name: "my path"
+            }
+    ]
+}
+~~~~
+
+... or a single map value with a certname key (equivalent to a bare string) ...
+
+~~~~hocon
+authorization: {
+    version: 1
+    rules: [
+            {
+                match-request: {...}
+                deny: {
+                    certname: node1
+                }
+                sort-order: 1
+                name: "my path"
+            }
+    ]
+}
+~~~~
+
 ... or as an array of values ...
 
 ~~~~hocon
@@ -526,7 +697,7 @@ authorization: {
     rules: [
             {
                 match-request: {...}
-                deny: [ node1, node2, node3 ]
+                deny: [ node1, node2, node3, {extensions:{ext_name:value}} ]
                 sort-order: 1
                 name: "my path"
             }
--- a/examples/ring_app/ring-example.conf
+++ b/examples/ring_app/ring-example.conf
@@ -30,6 +30,19 @@ authorization: {
             allow: "$1"
             name: "users allowed by backreference"
             sort-order: 500
-        }
+        },
+        {
+            match-request: {
+                path: "/hello/extension-allowed"
+                type: "path"
+            }
+            allow: {
+                extensions: {
+                    my_extension_shortname: "extensions value"
+                }
+            }
+            name: "Can authorize based on X.509 extensions."
+            sort-order: 500
+        },
     ]
 }
--- a/src/puppetlabs/trapperkeeper/authorization/acl.clj
+++ b/src/puppetlabs/trapperkeeper/authorization/acl.clj
@@ -3,6 +3,11 @@
 
 ;; Schemas
 
+(def OIDMap
+  "Mapping of string OIDs to shortname keywords. Used to update an incoming
+  request with a shortname -> value extensions map."
+  {schema/Str schema/Keyword})
+
 (def Extensions
   "Schema for representing SSL Extensions. Maps from a keyword shortname to a
   string value."
@@ -174,11 +179,14 @@
   *ONLY* a request with both :pp_env set to 'test' and :pp_image set to 'bad
   image' would be denied. If *any* request with :pp_env set to 'test' is to be
   denied, it needs a standalone deny rule."
-  [ace :- ACE
+  [oid-map :- OIDMap
+   ace :- ACE
    extensions :- Extensions]
   (let [match-key (fn [k]
                     (let [ace-value (get (:value ace) k)
-                          ext-value (get extensions k false)]
+                          ;; potentially translate from oid -> shortname
+                          k' (get oid-map (name k) k)
+                          ext-value (get extensions k' false)]
                       (if ext-value
                         (if (sequential? ace-value)
                           (some (partial = ext-value) ace-value)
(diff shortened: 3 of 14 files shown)


src/puppetlabs/trapperkeeper/authorization/acl.clj
```diff
@@ -56,123 +82,165 @@
...
 
-(schema/defn new-domain :- Entry
+(schema/defn ^:always-validate new-domain :- ACE
   "Creates a new ACE for a domain"
-  [type :- AuthType
-   pattern :- schema/Str]
+  [auth-type :- AuthType
+   {:keys [certname extensions]} :- ACEConfig]
   (cond
+    ;; CSR Attributes
```
camlow325: nit - should just refer to "extensions" here rather than CSR attributes, even though I suppose "extensions" would be obvious from the following code?

vilmibm: oops, forgot to do a final sweep for CSR in the code. Will do that now.

src/puppetlabs/trapperkeeper/authorization/acl.clj
```diff
@@ -56,123 +82,165 @@
...
-    (throw (Exception. (str "invalid domain pattern: " pattern)))))
+    (throw (Exception. (str "invalid domain value: " certname)))))
 
 ;; ACE matching
 
-(schema/defn match-name? :- schema/Bool
+(schema/defn ^:private match-domain? :- schema/Bool
   "Checks that name matches the given ace"
-  [{:keys [pattern match]} :- Entry
+  [{:keys [value match] :as ace} :- ACE
```
camlow325: Doesn't look like the `:as ace` gets used here.

src/puppetlabs/trapperkeeper/authorization/acl.clj
```diff
@@ -56,123 +82,165 @@
...
+  All of the keys in the ACE must appear in the extensions map and, if the value
+  for a key in the ACE is a list, at least one of the listed values must be set
+  in the incoming extensions map.
+
+  Note the behavior in the following scenario: If an ACE specifies
+  {:deny {:extensions {:pp_env 'test'
+                       :pp_image 'bad image'}}}
+
+  *ONLY* a request with both :pp_env set to 'test' and :pp_image set to 'bad
+  image' would be denied. If *any* request with :pp_role set to 'test' is to be
```
camlow325: Might make more sense in context here for this to say `pp_env` than `pp_role`.

src/puppetlabs/trapperkeeper/authorization/acl.clj
```diff
@@ -56,123 +82,165 @@
...
+                        false)))]
+    (every? match-key (keys (:value ace)))))
+
 (schema/defn match? :- schema/Bool
-  "Returns true if the given name matches the given ace"
-  [ace :- Entry
-   name :- schema/Str]
-  (match-name? ace name))
+  "Returns true if the given value matches the given ace"
+  [{:keys [match value] :as acl-ace} :- ACE
```
camlow325: Looks like `value` on this line and `incoming-ace` from the following line aren't used here.

src/puppetlabs/trapperkeeper/authorization/ring.clj
```diff
@@ -48,3 +49,15 @@
...
    name :- (schema/maybe schema/Str)]
   (assoc-in request [:authorization :name] (str name)))
+
+(schema/defn set-authorized-extensions :- Request
+  "Set the authorized extensions onto the request."
+  [request :- Request
+   extensions :- acl/Extensions]
+  (assoc-in request [:authorization :extensions] extensions))
+
+(schema/defn authorized-extensions :- {schema/Keyword String}
```
camlow325: Could use `acl/Extensions` schema for the return type here?

src/puppetlabs/trapperkeeper/authorization/ring.clj
```diff
@@ -48,3 +49,15 @@
...
+  "Get the authorized client name from the request."
```
camlow325: I think this should say something like "Get the authorized _extensions_..."

src/puppetlabs/trapperkeeper/authorization/ring_middleware.clj
```diff
@@ -17,6 +18,11 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Private
 
+(def OIDMap
+  "Mapping of string OIDs to shortname keywords. Used to update an incoming
+  request with a shortname -> value extensions map."
+  {String schema/Keyword})
```
camlow325: Better to use `schema/Str` than `String` here?  Not sure we've always been consistent about that anyway...

vilmibm: It's not consistent : / I see both in our code. I can switch this to schema/Str so it's at least consistently using the schema ns in this schema...

test/puppetlabs/trapperkeeper/authorization/testutils.clj
```diff
@@ -20,9 +20,13 @@
...
+  [cn & [extensions]]
+  (let [cacert (ssl/gen-self-signed-cert "my ca" 41 {:keylength 512})
+        opts {:keylength 512}
+        opts (if (nil? extensions)
+               opts
+               (assoc opts :extensions extensions))]
+    (:cert (ssl/gen-cert cn cacert 42 opts))))
 
 (def test-domain-cert (create-certificate "test.domain.org"))
 (def test-other-cert (create-certificate "www.other.org"))
```
MSLilah, line 32: I think there's a typo in the commit message for this commit

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -20,6 +20,32 @@
...
+
+(schema/defn validate-ace-config-map!
+  [config-value :- (schema/pred map? "map?")]
+  (when (:attrs config-value)
+    (schema/validate acl/CSRAttributes (:attrs config-value)))
+
+  (when (:certname config-value)
+    (schema/validate String config-value))
+
+  (when (or (not (or (:attrs config-value) (:certname config-value)))
```
MSLilah: Might be better to do this with a schema predicate? Then our error handling would be consistent across this function

vilmibm: This is pretty out of date code...

test/puppetlabs/trapperkeeper/authorization/testutils.clj
vilmibm, line 32: Yep, fixed.

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
MSLilah: Whoops, sorry, was reviewing by-commit. I'll switch to reviewing the full diff

test/puppetlabs/trapperkeeper/authorization/acl_test.clj
```diff
@@ -78,57 +79,120 @@
...
+        (is (!allowed? (challenge {:style "cutup" :publisher "knopf"})))))
+
+    (testing "with list matching"
+      (let [allowed? (partial acl/allowed?
+                              (acl/allow {:extensions {:style ["sonnet" "prose" "sestina"]
+                                                       :author "olds"}}))
+            !allowed? (comp not allowed?)]
+        (is (allowed? (challenge {:style "prose" :author "olds"})))
+        (is (!allowed? (challenge {:style "haiku" :author "olds"})))
+        (is (!allowed? (challenge {:stye "sestina" :author "burroughs"})))))
```
MSLilah: Might also be worth testing that `sonnet` and `sestina` are valid styles? Also looks like `stye` was used here instead of `style`, not sure if that's a typo or if that's intentional

MSLilah:
:see_no_evil: although I'm not an auth expert. I'm happy with this once @camlow325 is :+1:

test/puppetlabs/trapperkeeper/services/authorization/authorization_service_test.clj
```diff
@@ -5,7 +5,7 @@
     [puppetlabs.ssl-utils.core :as ssl-utils]
     [puppetlabs.trapperkeeper.app :refer [get-service]]
     [puppetlabs.trapperkeeper.authorization.rules :as rules]
-    [puppetlabs.trapperkeeper.authorization.testutils :refer :all]
+    [puppetlabs.trapperkeeper.authorization.testutils :as testutils :refer :all]
```
camlow325: Might be better to just do one or the other - the `:as testutils` or the `:refer :all`.

doc/authorization-config.md
```diff
@@ -451,7 +489,7 @@ authorization: {
     rules: [
             {
                 match-request: {...}
-                allow: [ node1, node2, node3 ]
+                allow: [ node1, node2, node3, {extensions:{ext_shortname: value}} ]
```
camlow325: Might be good to show an example with more than one extension in it - either here or as an additional one.

doc/authorization-config.md
```diff
@@ -494,6 +532,14 @@ One of the following forms may be used as the value for an `allow` entry:
  of `xyz.domain.org` and a request URL of
  `"http://my-host:8080/the/path/xyz"`, however, would not be considered a 
  match for the entry.
+ 
+ * A map with either a `certname` or `extensions` key. The former behaves
+   exactly as a standalone string. The latter is used to specify a set of ssl
```
camlow325: Maybe say X.509 certificate extensions here?

doc/authorization-config.md
```diff
@@ -494,6 +532,14 @@ One of the following forms may be used as the value for an `allow` entry:
...
+   extensions to match. Each key specified in the extensions map must appear in
+   the request and the values are matched exactly (no wildcarding). Extensions
+   in the request not explicitly listed here are ignored. Note that a given
+   extension map won't match if not *all* keys match. If you want an "or"
+   relationship between extensions, split them into different extension maps.
```
camlow325, line 542: Somewhere in this description, I'd mention wording like..

1) Each key included in the extension map must have a matching extension in the request certificate.
2) The value associated with each extension key may be a string or a list of strings.  If the extension value is a string, it must be an exact match for the corresponding extension's value in the request certificate in order for the entry to be a match.  If the extension value is a list, one or more of the items in the list must match the corresponding extension's value in the request certificate.

Where this is a little hard to explain, I think a couple of examples of what do / do not constitute a match would be helpful to include here.

examples/ring_app/ring-example.conf
```diff
@@ -30,6 +30,19 @@ authorization: {
...
+            match-request: {
+                path: "/hello/extension-allowed"
+                type: "path"
+            }
+            allow: {
+                extensions: {
+                    my_extension_shortname: "extensions value"
+                }
+            }
+            name: "users allowed by backreference"
```
camlow325: I think this will blow up since the names for each rule need to be unique.  Maybe make this "users allowed by matching extension" or something like that?

camlow325:
It would be good to update the `README.md` with some additional documentation about the new options map - and associated `oid-map` entry - that the `wrap-with-authorization-check` and `authorization-check` service functions support.

test/puppetlabs/trapperkeeper/services/authorization/authorization_service_test.clj
vilmibm: The file already had a :refer :all which infuriated me so I added an :as for the imports I used. I can audit the file and make it all relative and take the :all out.

camlow325:
It's certainly possible that I was doing something wrong, but I was a bit surprised with the result I got for the following scenario...

1) Using the ring-example in the project, I modified the 'extension-allowed' rule to have the following:

``` hocon
    {
       match-request: {
           path: "/hello/extension-allowed"
           type: "path"
       }
       allow: {
           extensions: {
                "2.16.840.1.113730.1.13": "Puppet Ruby/OpenSSL Internal Certificate"
           }
        }
        name: "users allowed by extension"
        sort-order: 500
   }
```

2) In the ring handler registration for the `hello-service` in `ring_app.clj`, I had:

``` clj
(add-ring-handler
   (wrap-with-authorization-check
      (fn [_]
        {:status  200
         :headers {"Content-Type" "text/plain"}
         :body    "Hello, World!"})
      {:oid-map {"2.16.840.1.113730.1.13" :netscape_comment}})
      "/hello")
```

When I made a curl request using the `localhost` cert from the project...

```
curl --ipv4 "https://localhost:8081/hello/extension-allowed" --cacert ./examples/ring_app/ssl/certs/ca.pem --cert ./examples/ring_app/ssl/certs/localhost.pem --key ./examples/ring_app/ssl/private_keys/localhost.pem
```

... I got a 403 / Forbidden Request:

```
Forbidden request: localhost(127.0.0.1) access to /hello/extension-allowed (method :get) (authenticated: true) denied by rule 'users allowed by extension'
```

The request was allowed when I used the short name in my rule definition, though:

``` hocon
   allow: {
      extensions: {
         "netscape_comment": "Puppet Ruby/OpenSSL Internal Certificate"
      }
   }
```

I assume this means that one would have to use the short name in the rule definition if a short name -> OID mapping exists although I think it might be friendlier to allow the OID to still be used in the rule definition even if a mapping doesn't exist.  I don't have a strong opinion on this, though, so if the UX team and others are good with this, then I am too.

vilmibm:
@camlow325 Oops. Yeah, though we discussed it, I forgot to support the OID in the config file :(. I will add support for this tomorrow (unless you think it should be a follow up ticket).

camlow325:
Overall big :+1: on this work.  Great job with continued refactorings / cleanups along with the work for the new feature.  The implementation to support the extensions looks pretty clean to me.  I played around with the extension matching behavior a bit with the ring-example app.  Seemed to be working the way I had expected it do.

I left a few cosmetic comments on the PR.  The main items that I think might warrant some follow-up are:

1) The case where matching doesn't work when an OID might be used for the extension in the rule definition but the short name is used in the supplied mapping file.

2) A little more detail in the docs, including examples.

I think we should also think through exactly what we want the matching behavior to be for cases where the extension's value is not just a simple ASN.1 'String' type.  The primary one that we'll need to consider is the SubjectAltName, for which [TK-352](https://tickets.puppetlabs.com/browse/TK-352) was carved off.  I'm good with revisiting this behavior more generally in that follow-on ticket.

When whatever of my comments above might be addressed, I'm good with someone merging this.

src/puppetlabs/trapperkeeper/authorization/acl.clj
```diff
@@ -188,18 +196,21 @@
...
-    (= :extensions match)
-    (if (nil? extensions)
-      false
-      (match-extensions? acl-ace extensions))
-
-    :else
-    (if (nil? certname)
-      false
-      (match-domain? acl-ace certname))))
+  ([acl-ace incoming-ace]
```
MSLilah: Might be worth keeping the schema on these parameters

MSLilah:
I'm :+1: on this

MSLilah:
@KevinCorcoran Did you want to review this or should I go ahead and merge?

KevinCorcoran:
@fpringvaldsen If you and @nathanielksmith think this is ready for merge now, :shipit:

vilmibm:
i believe this is ready for merge and has been for a few days :)

KevinCorcoran:
@camlow325 please feel free to follow-up if you have any post-merge feedback.

doc/authorization-config.md
```diff
@@ -494,6 +532,101 @@ One of the following forms may be used as the value for an `allow` entry:
...
+   ~~~~clojure
+   {:role "console"
+    :env "appgroup1"
+    ;; irrelevant keys elided
+   }
+   ~~~~
+
+   ~~~~clojure
+   {:role "console"
+    :env "appgroup1"
```
camlow325, line 626: Looks like this example is exactly the same as the last one?

examples/ring_app/ring-example.conf
```diff
@@ -30,6 +30,19 @@ authorization: {
...
-        }
+        },
+        {
+            match-request: {
+                path: "/hello/extension-allowed"
+                type: "path"
+            }
+            allow: {
+                extensions: {
+                    my_extension_shortname: "extensions value"
```
camlow325, line 41: Not a big deal either way, but I find it useful with these examples to have some ready-made certificates in the repo that I can play around with against the rules in the ring example.  For my testing, I just used the existing cert in the repo and experimented with two different allow/deny rules for the 'netscape comment' (2.16.840.1.113730.1.13) extension, since that extension already existed in the dummy cert in the repo.  Not worth adding anything else at this point, I think, but maybe something to keep in mind if we're making changes around this again in the future?

camlow325:
Retroactive :+1: @nathanielksmith++

PR #36, opened 2016-04-19
(maint) Check schema of just the certname

haus:
Previously the schema check for an ace-config entry would attempt to
validate that a map and not the value in the map was a string. As a map
is never a string, trying to use a certname entry in an allow/deny map
would not work. This commit corrects that to use the value and not the
map itself.

### DIFF
--- a/src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
@@ -142,9 +142,16 @@
         get-br-list (fn [match-str]
                       (map #(Integer/parseInt (second %))
                            (re-seq #"\$(\d+)" match-str)))
+        get-certname-list (fn [rule]
+                            (cond
+                              (string? rule) (list rule)
+                              (map? rule) (list (get rule :certname ""))
+                              (seq? rule) (for [x rule]
+                                                 (if (map? x) (get x :certname "") x))
+                              :else (list "")))
         largest-br (fn [match]
                      (let [br-list (mapcat get-br-list
-                                           (flatten [(or match "")]))]
+                                           (get-certname-list match))]
                        (when (> (count br-list) 0)
                          (apply max br-list))))
         largest-allow-br (or (largest-br (:allow rule)) 0)
--- a/src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
@@ -32,7 +32,7 @@
                (:extensions config-value))))
 
   (when (:certname config-value)
-    (when-not (nil? (schema/check schema/Str config-value))
+    (when-not (nil? (schema/check schema/Str (:certname config-value)))
       (reject! (format "certname key should map to a string; got '%s'" (:certname config-value)))))
 
   (when (or (not (or (:extensions config-value) (:certname config-value)))
--- a/src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
@@ -146,7 +146,7 @@
                             (cond
                               (string? rule) (list rule)
                               (map? rule) (list (get rule :certname ""))
-                              (seq? rule) (for [x rule]
+                              (sequential? rule) (for [x rule]
                                                  (if (map? x) (get x :certname "") x))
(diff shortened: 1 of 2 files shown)


haus:
I'm looking to see if there's an obvious test to add here.

haus:
There is something else funky going on. Please hold.

haus:
Have some test and other fixes in progress. Will push out tomorrow.

haus:
okay i think this is ready for review.

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -142,9 +142,16 @@
         get-br-list (fn [match-str]
                       (map #(Integer/parseInt (second %))
                            (re-seq #"\$(\d+)" match-str)))
+        get-certname-list (fn [rule]
```
haus, line 145: I'm contemplating moving this to a proper outer function, mainly so I could write a couple of unit tests for it.

vilmibm:
:+1:

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
vilmibm, line 145: I feel like it's being sufficiently tested as-is...

haus, line 145: That's fair

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -142,9 +142,16 @@
...
+                            (cond
+                              (string? rule) (list rule)
+                              (map? rule) (list (get rule :certname ""))
+                              (seq? rule) (for [x rule]
```
camlow325: I don't think `seq?` works here.  If I change this to `sequential?`, though, it seems to work for the `vector` type that this can come in as from HOCON.

haus: good catch. i'll push out an update.

haus: updated

vilmibm: good catch. those two are always so hard to keep straight. rhickey--

test/puppetlabs/trapperkeeper/services/authorization/authorization_core_test.clj
```diff
@@ -13,8 +13,12 @@
                       :sort-order 500 :name "base-regex-auth"})
 (def allow-single {:allow "www.domain.org"})
 (def allow-list {:allow ["*.domain.org" "*.test.com"]})
+(def allow-list-of-maps {:allow ["foo.bar.com", {:extensions {:foo "bar"}}]})
+(def allow-map {:allow {:certname "bald.fish.com"}})
 (def deny-single {:deny "bald.guy.com"})
 (def deny-list {:deny ["bald.eagle.com" "bald.bull.com"]})
+(def deny-map {:deny {:extensions {:thing "bald.wolf.com"}}})
+(def deny-list-of-maps {:deny ["foo.bar.com", {:certname "bar.nate.wolfe"}]})
```
camlow325, line 21: We should just hardcode a "nate.wolfe" deny rule in tk-auth.  :smile:

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
camlow325, line 145: Having as an outer function for readability might be a good thing since the amount of code in the embedded function is getting a bit long but I'm okay with this as it is.

Since we obviously didn't have any tests that cover out of bounds backreferences for the more complex ACEs that we can do now, it may be worth at least adding a couple of higher-level tests like these ones - https://github.com/puppetlabs/trapperkeeper-authorization/blob/0.6.0/test/puppetlabs/trapperkeeper/services/authorization/authorization_core_test.clj#L215-L250 - which have `certname` in an ACE map entry.

camlow325:
:+1: Awesome catch.  Just had the couple of comments about `seq?` vs. `sequential?` - which it looks like you already fixed - and possibly adding a couple of extra out of bounds backreference tests for the new code added to `validate-regex-backreferences!`.

camlow325:
:+1: :ship:

PR #37, opened 2016-04-27
(TK-352) support subjectAlternativeName

vilmibm:
This PR adds awareness to the subjectAlternativeName certificate
extension. Specifically, that means that we now:
- extract a list of DNS alt names from subjectAlternativeName
- consider a request to match a rule if its certname appears in that dns
  alt name list
- consider a request to match a rule if that rule specifics a
  subjectAlternativeName value that appears in that dns alt name list
- have a default OID map that translates subjectAlternativeName

This PR also adds some additional docs to this project.


vilmibm:
@camlow325 would love a review from you on this. @cprice404 also, if you have the time.

doc/readme.md
```diff
@@ -0,0 +1,16 @@
+# Trapperkeeper Authorization
```
camlow325: I'd be inclined to not include this new file and instead work any new content you think would be useful about extensions into the existing top-level `README.md` file in the repo.  Sound reasonable?

src/puppetlabs/trapperkeeper/authorization/acl.clj
```diff
@@ -11,7 +13,8 @@
 (def Extensions
   "Schema for representing SSL Extensions. Maps from a keyword shortname to a
   string value."
-  {schema/Keyword schema/Str})
+  {schema/Keyword schema/Str
+   (schema/optional-key :subjectAlternativeName) [schema/Str]})
```
camlow325: We'd talked about this already, but I think it would be good to get a consult from someone on the docs team about what we want to call this exactly.  Where the other standard Puppet ones would have names like `pp_uuid`, maybe something like `subject_alternative_name` -- or just shorten to `subject_alt_name` like the X.509 RFC does (https://tools.ietf.org/html/rfc5280#section-4.2.1.6) -- would be more consistent?

camlow325: I haven't read far enough into this yet to know if this applies, but it seems like it would be good to allow the subj-alt-name value in the rule configuration to support both a list of strings or just a single string - given that we support the use of a single string in other parts of a rule.  If the value from the rule in the TK configuration is coerced into a vector before it is evaluated by this schema, my comment would be irrelevant.

src/puppetlabs/trapperkeeper/authorization/acl.clj
```diff
@@ -182,15 +191,17 @@
   [oid-map :- OIDMap
    ace :- ACE
    extensions :- Extensions]
-  (let [match-key (fn [k]
+  (let [oid-map (merge default-oid-map oid-map)
```
camlow325: As a minor style thing, I'd tend to prefer using a different name than `oid-map` for the bound variable - even if just `oid-map'`.  Same for the redefined `ace-value` variable below.

src/puppetlabs/trapperkeeper/authorization/ring_middleware.clj
```diff
@@ -209,11 +209,14 @@
    allow-header-cert-info :- schema/Bool
    oid-map :- acl/OIDMap]
   (if-let [cert (request->cert request allow-header-cert-info)]
-    (let [extensions (ssl-utils/get-extensions cert)
+    (let [oid-map (merge acl/default-oid-map oid-map)
```
camlow325, line 208: Since there are already a couple of places where the `default-oid-map` is getting merged with the custom `oid-map`, I wonder if it would be better to merge them in one place (at a higher-level) and pass the merged map on down.  Maybe in `ring-middleware/authorization-check?`

src/puppetlabs/trapperkeeper/authorization/ring_middleware.clj
```diff
@@ -209,11 +209,14 @@
...
+          extensions (ssl-utils/get-extensions cert)
           translate-oids (fn [out extension]
                            (let [oid-key (get oid-map (:oid extension)
                                               (keyword (:oid extension)))
-                                 value (str (:value extension))]
+                                 value (if (= :subjectAlternativeName oid-key)
+                                         (ssl-utils/get-subject-dns-alt-names cert)
```
camlow325: With respect to the matching that we're doing where a 'certname' value for an ACE could match a Subject CN or a dNSName entry in the Subject Alternative Name extension, I think this makes sense.

From https://tools.ietf.org/html/rfc2818#section-3.1:

```
If a subjectAltName extension of type dNSName is present, that MUST
be used as the identity. Otherwise, the (most specific) Common Name
field in the Subject field of the certificate MUST be used. Although
the use of the Common Name is existing practice, it is deprecated and
Certification Authorities are encouraged to use the dNSName instead.
```

Arguably the above could be interpreted to mean we should only match from a dNSName entry - and not the Subject CN - if both are present in a certificate.  I don't think we can make that leap, though, for compatibility reasons, so I'm good with what you have here from that perspective.

However, I'm thinking it may be better for an ACE that expresses a SubjectAltName constraint to not forceably limit the matches to dNSName entries only.  For example, I think it might be better to allow for it to match per any of the available GeneralName types - dNSName, iPAddress, ...

From that perspective, it might actually be better for rules with SubjAltName ACEs to be able to have the match include the GeneralName type.  So, for example, you could express a rule in your TK config that would have something like:

```
allow: {
  extensions: {
    subject_alt_name: 
       {  dns_name: [ "myhost1", "myhost2" ],
          ip_addr: [ "1.2.3.4" ]
       }
    }
  }
```

I'd be fine even with not implementing support for anything but "dns_name" as a GeneralName type for now.  If we require the names to be qualified with "dns_name" at the rule level for now, though, that would give us the ability to expand to support other GeneralName types later when/if needed.

Does this make sense?

test/puppetlabs/trapperkeeper/services/authorization/authorization_service_test.clj
```diff
@@ -451,3 +451,66 @@
...
+                (is (-> (get-service app :PlumbingService)
+                        (call-authorization-check req auth-opts)
+                        rules/authorized?))))
+            (testing "with deny rule"
+              (let [req (testutils/request "/puppet/v4/catalog" :get "127.0.0.1" deniable-cert)]
+                (is (not (-> (get-service app :PlumbingService)
+                             (call-authorization-check req auth-opts)
+                             rules/authorized?))))))))
+
+      (testing "as a certname match"
```
camlow325, line 517: These tests look good.  I think it might be good also to have at least one more flavor of domain rule - maybe a regex? - to show that the other domain rule type still matches a subjaltname from the certificate as expected?

doc/extensions.md
```diff
@@ -0,0 +1,24 @@
+# Working with x.509 Extensions in tk-auth
```
camlow325, line 1: Where the `authorization-config.md` doc is starting to get pretty long, having a separate document that just talks about the extensions - especially as we may intend to add more standard ones in the future - seems like a good idea.  I do think it would be good to link from `authorization-config.md` back to this doc if we intend to do that, though.  Maybe we could embed links from here - https://github.com/puppetlabs/trapperkeeper-authorization/blob/0.6.0/doc/authorization-config.md#allow - and here - https://github.com/puppetlabs/trapperkeeper-authorization/blob/0.6.0/doc/authorization-config.md#deny?

camlow325:
Overall, :chart_with_upwards_trend: on this.  Looking good!  Glad to see how little code needed to be changed.

I played around with this a bit with customizations to the config in the ring-example and curl requests to a bootstrapped server.  Everything appeared to work as I'd expected.

The biggest questions I have at this point are related to naming and the layering of dns_name vs. other types of SubjAltNames, mentioned in my comment here - https://github.com/puppetlabs/trapperkeeper-authorization/pull/37#discussion_r61524666.  @cprice404 - no problem if you don't have time to review this in its entirety but I would at least appreciate your take on my comment there.  If it would be easier for the three of us to talk about this IRL to clarity the comment first, definitely open to that.

doc/readme.md
vilmibm: I think having a page like this is really helpful. I noticed us doing it in tk-status and think it helps a newcomer figure walk through the docs.

However, if there's a reason you think an index page is problematic I def want to hear it.

camlow325: @nathanielksmith I don't have a big problem with having the separate file.  I'm probably just used to not looking for a readme.md file in the "doc" folder since there's already one in the base directory.  Might be good to get someone else's opinion on this - especially someone with little/no prior experience with the tk-auth documentation.  Ultimately, if you want to keep this file in, though, I'm good with it.

jpinsonault: Maybe have a pointer from the top level readme to these docs?

KevinCorcoran: I don't have any problem with the existence of this file.  If @nathanielksmith has found it helpful in other repos, I don't see any reason to remove it.  However, I share some of @camlow325's concerns about either duplicating content here or getting into a split-brain situation where it's not clear if things should live in this file or the top-level README.md.  So I +1 @jpinsonault's comment about linking to the top-level README as well as perhaps deleting some of the text in this file which explains what this project is/isn't, etc. (as that should be covered in the top-level README) and leave this doc as just a sort of "meta-doc".

vilmibm: Works for me. I can move any info here not in the readme to the readme and then maybe rename this file "index.md". That might help clarify the distinction.

src/puppetlabs/trapperkeeper/authorization/acl.clj
vilmibm: 1) agreed that subject_alt_name is way better; renamed
2) that schema is for what comes out of a cert for subjectAlternativeName, which is always going to be a list of strings. The ACEConfig schema is for the configuration side and remains unchanged.

src/puppetlabs/trapperkeeper/authorization/acl.clj
vilmibm: Fixed.

src/puppetlabs/trapperkeeper/authorization/ring_middleware.clj
vilmibm, line 208: I thought of that... My concern was that the functions being called (like allowed) can totally be called in isolation outside of the middleware and as such I wanted the defaults applied at the lowest levels, not the highest level. I found the two lowest-level uses of the oid-map and did the merging there to make sure there wasn't a possible call site that would miss the defaults.

Does that make sense / seem reasonable? Obviously the downside is if another low-level use crops up, someone has to know to do this, so I'm not 100% convinced I'm right here.

src/puppetlabs/trapperkeeper/authorization/ring_middleware.clj
vilmibm: Makes perfect sense, and I was wondering if I should be doing this exact thing. Thanks for bringing it up.

This will take me a bit of time to add since it means special case parsing on the config side as well as the request side. I'll try to have the PR updated by eod monday.

test/puppetlabs/trapperkeeper/services/authorization/authorization_service_test.clj
vilmibm, line 517: Agreed; but I'll wait on this change until I've got the above stuff done since that'll require new/different tests.

src/puppetlabs/trapperkeeper/authorization/ring_middleware.clj
camlow325, line 208: @nathanielksmith Yeah, I understand where you're coming from on that.  I still think I'd be more nervous about missing the application of the default oids if we have to do it in multiple places in various utility code rather than just in one place at the top of the incoming injection from the middleware API functions.

@KevinCorcoran - what do you think about this?  Would you have a specific preference?

KevinCorcoran, line 208: I don't have a preference.

camlow325, line 208: k, well I'll defer to whatever @nathanielksmith wants to do on this, then.

vilmibm, line 208: Since the merge is idempotent I could do it at both levels ^_^

test/puppetlabs/trapperkeeper/services/authorization/authorization_service_test.clj
vilmibm, line 517: This is the last thing I have to do; naturally, there's a bug in my handling of wildcarded allow/deny domains :) will push up a hopefully final version once i have that fixed.

vilmibm:
@camlow325 All feedback addressed.

camlow325:
👍 Looks really good.  The only other suggestion I'd have is that we put a little more info into the subj_alt_name extension docs to show the different kinds of General Names that you can use in an ACE value.  Functionality all looks good to me, though!

camlow325:
I guess I did have one other cosmetic comment.  When doing some functional testing with this, I found it a little confusing that the name of the "subject_alt_name" OID short name had underscores but the keys in the subordinate map had dashes - e.g., "dns-name".  I think it might be a seem a little more consistent for end users to be able to use underscores for both.

doc/extensions.md
```diff
@@ -0,0 +1,24 @@
...
+
+When a certificate is signed, one can include key value data using x.509
```
KevinCorcoran: I'd reword "key value data" as "arbitrary key-value pairs".

src/puppetlabs/trapperkeeper/authorization/acl.clj
```diff
@@ -11,13 +13,36 @@
 (def Extensions
   "Schema for representing SSL Extensions. Maps from a keyword shortname to a
   string value."
```
KevinCorcoran: docstring could stand to be updated

vilmibm:
@camlow325 @KevinCorcoran updated.

test/puppetlabs/trapperkeeper/services/authorization/authorization_service_test.clj
justinstoller, line 517: I assume the resolution to this is what you referred to in https://github.com/puppetlabs/trapperkeeper-authorization/pull/37#issuecomment-219502032

vilmibm, line 517: yep! there wasn't a bug after all, i just mistyped a string in a test.

justinstoller:
It all looks like it does what it says on the tin. The docs are a great addition, and all the tests pass when merged into master locally. Has/should someone run this at the acceptance (using this with puppet) level?

👍 If Jeremy wants to merge!

vilmibm:
@justinstoller I haven't run at the acceptance test level since this is not intend to interact with puppet in any meaningful way--it's an additive change to how we process and enforce tk-auth rules.

camlow325:
@justinstoller I played around with this a bit at closer to an "acceptance" level by building a tk-auth jar with these changes and updating a local branch of Puppet Server to use it.  I used some custom tk-auth rules to see that certname matches against a subj-alt-name / dns-name entry from my SSL client certificate was sufficient to match (allowing) the request and that I could successfully do similar matches with a rule using "extensions: { subject-alt-name: { dns-name: ... } }" in an ACE.  This all looks good to me.

So, 👍 from me.  Sounds like others are good with this as well so I'm going to go ahead and merge this...

PR #38, opened 2016-05-03
Consolidate common middleware and helpers

justinstoller:
As part of TK-214, we extended the middleware and middleware helpers in
the shared library ring-middleware.

This refactors tk-auth to use the bad-request helpers and invalid data
middleware.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,5 +1,5 @@
-(def ks-version "1.0.0")
-(def tk-version "1.1.1")
+(def ks-version "1.3.0")
+(def tk-version "1.4.0")
 (def tk-jetty-version "1.4.1")
 
 (defproject puppetlabs/trapperkeeper-authorization "0.5.1-SNAPSHOT"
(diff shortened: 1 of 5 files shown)


justinstoller:
This will need to be updated with a newer version on `puppetlabs/ring-middleware` once this PR https://github.com/puppetlabs/ring-middleware/pull/28 has been merged and released. This PR should allow anyone who wants to run the tests via a checkout of the mentioned PR.

KevinCorcoran:
🙈

justinstoller:
I think this is ready to go

project.clj
```diff
@@ -13,21 +13,28 @@
   ;; requires lein 2.2.0+.
   :pedantic? :abort
 
+  :exclusions [org.clojure/clojure]
```
KevinCorcoran: This looks suspect ...

test/puppetlabs/trapperkeeper/services/authorization/authorization_service_test.clj
```diff
@@ -290,7 +290,7 @@
                   (update-in [:headers] assoc "x-client-cert" "NOCERTS"))
           {:keys [status body]} (app req)]
       (is (= status 400))
-      (is (= body "No certs found in PEM read from x-client-cert")))))
+      (is (re-matches #".*No certs found in PEM read from x-client-cert.*" body)))))
```
KevinCorcoran: Why was this change necessary?

project.clj
justinstoller: This is probably me not knowing what I'm doing.

This repo has a dep on Clojure 1.6, while all of our other libraries that I touched in this process depend on 1.7. Changing the version of Clojure that a library depended on seemed like a much bigger deal/possible api breaker than bumping any individual library, so instead of bumping this project to 1.7 I added this `:exclusions` like I had seen in [tk-status](https://github.com/puppetlabs/trapperkeeper-status/blob/master/project.clj#L12).

If bumping the version of Clojure is something we can do in a minor refactor then I'll do that?

KevinCorcoran: Was this technically necessary or did it just seem like A Good Thing To Do?

TBH I'm not sure what the best approach is here.  I didn't realize that tk-status was doing this as well.  After some more thought it seems reasonable.

justinstoller: lein pedantic will abort because `ring-middleware`, `http-client` (a transitive), and `clj-time` require Clojure 1.7. It then recommends doing an exclusion on those libraries for Clojure.

I don't know why the explicit reference to Clojure doesn't already override them, but either bumping the Clojure dependency here, adding exclusions to those libraries, or adding a global exclusion is required to satisfy lein pedantic.

I went with the global exclusion because I thought bumping the Clojure version might be considered a breaking change and I had seen another project do it. I don't know if that makes it right though.

KevinCorcoran: If upstream projects are already depending on 1.7 and _not_ excluding it, that seems to be what should be done here as well, for the sake of consistency.  I don't think that's a breaking change.

test/puppetlabs/trapperkeeper/services/authorization/authorization_service_test.clj
justinstoller: This doesn't look necessary.

Thinking back on it I believe I updated the [middleware here](https://github.com/puppetlabs/trapperkeeper-authorization/pull/38/files#diff-983d1c3afdb0656721ce0c5b3f4f8049R45) to use the default behavior of `wrap-bad-request` (output json), updated the test to check that the content was in there without concern for its format. Realized I shouldn't be changing the format if at all possible, updated the middleware to output plain text, but forgot to revert the test change.

I can revert this test change, or add an assertion that the content type is also in plain text.

rlinehan:
👍 lgtm

test/puppetlabs/trapperkeeper/services/authorization/authorization_service_test.clj
```diff
@@ -288,9 +288,10 @@
                   (assoc :uri "/puppet/v3/environments")
                   (assoc :body "Hello World!")
                   (update-in [:headers] assoc "x-client-cert" "NOCERTS"))
-          {:keys [status body]} (app req)]
+          {:keys [status body headers]} (app req)]
       (is (= status 400))
-      (is (= body "No certs found in PEM read from x-client-cert")))))
+      (is (re-matches #".*No certs found in PEM read from x-client-cert.*" body))
```
KevinCorcoran: still think this could just be `=` instead of `re-matches`

justinstoller: Updated in addd1b7

PR #39, opened 2016-05-05
(TK-360) Remove IP address from responses to forbidden reqs

rlinehan:
Previously, an unauthenticated client could make a request to a route
protected by TK auth and get back an error message including the IP address
and rule description. This is more information than is probably good to
provide over http, especially since it could potentially include internal IPs
of proxy servers, depending on how those are configured. This commit makes it
so that the response includes minimal information but the full information is
logged server-side.


jpinsonault:
👍 Looks good to me

camlow325:
👍

PR #47, opened 2016-06-15
(TK-381) Clean up tk-auth deps

justinstoller:
This PR:
- Moves the dependency conflicts section to the top of the dependency list per convention.
- Bumps us to the latest and greatest of kitchensink
- Removes unneeded dependencies:
  - `org.clojure/tools.cli`
  - `me.raynes/fs`
  - `puppetlabs/typesafe-config`

(it looks like at one point we planned on having a cli interface and separate config for tk-auth)

I'm curious if anyone else knows of a reason that these would be necessary?

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,4 +1,4 @@
-(def ks-version "1.3.0")
+(def ks-version "1.3.1")
 (def tk-version "1.4.0")
 (def tk-jetty-version "1.4.1")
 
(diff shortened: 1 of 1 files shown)


rlinehan:
It looks like those dependencies were added by Brice when he was working on this, so it's entirely possible that his initial plan was for some CLI tools, or they were maybe just copied from another library. I don't think there's any reason why they are currently necessary.

👍

KevinCorcoran:
I'm pretty sure they were just there to resolve version conflicts in transitive dependencies.

PR #49, opened 2016-09-15
(200) Add MAINTAINERS

cprice404:


camlow325:
👍 Seems fine.  Dunno if we want to put a link to the maintainers doc from the README, like we did for some other repos where we've done this already.

justinstoller:
👍 I'd be for just merging. /cc @haus @KevinCorcoran

PR #51, opened 2016-10-14
(SERVER-1554) externalize strings

jpinsonault:


src/puppetlabs/trapperkeeper/authorization/ring_middleware.clj
```diff
@@ -59,18 +58,10 @@
...
-  ([common-name empty-message-format empty-message-arg1 empty-message-arg2]
-   (if (empty? common-name)
-     (log/warnf (str empty-message-format
-                     "  Treating client as 'unauthenticated'.")
-                empty-message-arg1
-                empty-message-arg2))
-   common-name))
+  [common-name empty-message]
+  (if (empty? common-name)
+    (log/warnf (format "%s  %s" empty-message (trs "Treating client as ''unauthenticated''."))))
```
camlow325, line 63: I'm not sure if we're supposed to be concatenating the result of one or more trs/tru calls together, e.g., via a format string, since it might create issues around translation.  Instead, I'd been thinking we needed to have the full English text version of the format constructed at the point where translation would need to be done.  So maybe here, even though it seems a bit redundant, we'd need to construct strings like the following in `request->name*` and pass those through to be logged verbatim in `warn-for-empty-common-name`?

``` clj
(trs "CN could not be found in certificate DN: {0}.  Treating client as ''unauthenticated''." cert-dn)
(trs "CN could not be found in RFC 2253 DN provided by HTTP header: ''{0}'': {1}.  Treating client as ''unauthenticated''." header-dn-name header-dn-val)
```

src/puppetlabs/trapperkeeper/authorization/rules.clj
```diff
@@ -142,26 +143,27 @@
...
   (let [from (requestor request)
         path (:uri request)
         method (:request-method request)
         authenticated? (true? (ring/authorized-authenticated request))]
-    (str "Forbidden request: " from " access to " path " (method " method ")"
-         (if-let [file (:file rule)] (str " at " file ":" (:line rule)))
-         (format " (authenticated: %s) " authenticated?)
-         (format "denied by rule '%s'." (:name rule)))))
+    (if-let [file (:file rule)]
+      (tru "Forbidden request: {0} access to {1} (method {2}) at {3}:{4} (authenticated: {5}) denied by rule ''{6}''."
```
camlow325: I think this and the next one should be `trs` instead of `tru`.  Both would be logged to the system log file - not returned in the response to the web request.

src/puppetlabs/trapperkeeper/authorization/rules.clj
```diff
@@ -190,11 +192,11 @@
     resp-reason :- schema/Str]
    (if rule
      (log/tracef
```
camlow325: Nit - I guess some of these `log/tracef` calls could be changed into `log/trace` calls now - since presumably the format parameters are handled by the `trs` macro and not passed along to the log function call?

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -159,90 +157,85 @@
...
-               " does not contain a '" (name k) "' key.")))
+      (reject! (trs "The authorization rule specified as {0} does not contain a ''{1}'' key."
+                    (pprint-rule rule) (name k)))))
   (when (or (not (integer? (:sort-order rule)))
             (< (:sort-order rule) 1)
             (> (:sort-order rule) 999))
-    (reject! "The sort-order set in the authorization rule specified as "
-             (pprint-rule rule) " is invalid. It should be a number "
-             "from 1 to 999."))
+    (reject! (trs "The sort-order set in the authorization rule specified as {0} is invalid. "
```
camlow325, line 183: I think it might be better here to collapse the two strings into a single line, like:

``` clj
(trs "The sort-order set in the authorization rule specified as {0} is invalid. It should be a number from 1 to 999."
  (pprint-rule rule))
```

Same comment on several of the other `reject!` calls in this namespace.

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -159,90 +157,85 @@
...
-        (reject! "The " (name type) " list in the rule specified as "
-                 (pprint-rule rule)
-                 " contains one or more aces that are not maps or strings."))
+        (reject! (trs "The {0} list in the rule specified as {1} contains one or more ACEs that are not maps or strings."
+                      (name type) (pprint-rule rule))))
       (when-not (or (map? aces) (string? aces))
-        (reject! "The ACE '" aces "' in the '" (name type) "' field of "
-                 "the rule specified as " (pprint-rule rule) " is invalid. "
-                 "It should be a string or a map with keys :extensions or :certname."))))
+        (reject! (trs "The ACE ''{0}'' in the ''{1}'' field of the rule specified as {2} is invalid. "
```
camlow325, line 218: I wonder if it would be better here to have `name` and `type` be different parameters, like "... in the ''{1} {2}'' field of the rule...".

src/puppetlabs/trapperkeeper/services/authorization/authorization_service.clj
```diff
@@ -19,7 +20,7 @@
    [this context]
    (let [config (get-in-config [:authorization])
          rules (-> config validate-auth-config! :rules transform-config)]
-     (log/debug "Transformed auth.conf rules:\n" (ks/pprint-to-string rules))
+     (log/debug (trs "Transformed auth.conf rules:") "\n" (ks/pprint-to-string rules))
```
camlow325: Should this have a '{0}' at the end of the format string - for substitution of the `(ks/pprint-to-string rules)` form?

camlow325:
Initially, I was thinking it might make sense to have a call to the i18n `locale-negotiator` wrapped around the `wrap-with-authorization-check` middleware.  But I suppose if we assume that whatever downstream service that injects its handler into `wrap-with-authorization-check` would do its own higher-level wrapping of the `locale-negotiator` that the extra wrapping at this level would be unnecessary - and maybe even cause bad side effects if the `locale-negotiator` middleware can't safely be sourced twice from the same handler stack.  So I think I'm good with not having it included here.  Just wanted to mention that in the review to see if others were thinking the same as I am here.

camlow325 (commented):
I'm generally 👍 on this.  I left a few comments about places where I think it might be better to combine multiple strings into a single trs/tru rather than having them be separate and potentially concatenated together after the fact.  TBH, I'm not sure if my suggestions here make sense / are strictly necessary since I don't really grok how all the translation will be done.  It just seemed like it was best to include the full message text in a single string so that and subtle variations on translation could be considered.

src/puppetlabs/trapperkeeper/authorization/rules.clj
jpinsonault: Oh yeah good call

src/puppetlabs/trapperkeeper/authorization/ring_middleware.clj
jpinsonault: I _think_ as far as the output is concerned it will work fine so long as we're translating and then concatenating and not concatenating and then translating. But yeah for the translators having the extra context would probably be better. Since there's only a few called to this I think it'd be fine to be a little redundant

src/puppetlabs/trapperkeeper/services/authorization/authorization_service.clj
jpinsonault: Yes! Thank you for catching that

jpinsonault: Wait, actually doesn't `log/debug` take any number of params and concatenate them? Either way one trs call with everything would read easier

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
jpinsonault: Yeah, I should probably be nice to the translators and give them the bigger context to work with

src/puppetlabs/trapperkeeper/services/authorization/authorization_service.clj
camlow325: Yeah, I think it does.  I was just thinking we wanted to always have the full body of the message wrapped inside of the trs format string, not that it would necessarily influence how this particular string might be translated.

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
jpinsonault: I'm not sure I understand what you mean. `name` is just a function here right?

camlow325: D'oh.  Somehow my eyes blocked out the parens around those and I wasn't looking that closely.  Nothing to see here - never mind.  🚶

src/puppetlabs/trapperkeeper/authorization/rules.clj
jpinsonault: Ah yep, you're right

jpinsonault:
Ok I responded to all the feedback that wasn't about un-splitting up the strings in light of Larissa's comments

cprice404 (approved):

cprice404:
@camlow325 if you had anything else on this feel free to suggest additional changes.

PR #53, opened 2017-01-31
(TK-430) Migrate to clj-parent 0.3.3 and bump i18n to 0.6.0

camlow325:
This commit migrates tk-auth to using clj-parent version 0.3.3 for its
dependencies.  This commit also updates the i18n plugin dependency from
0.4.3 to 0.6.0 for compatibility with the version of i18n being
inherited from clj-parent.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,38 +1,30 @@
-(def ks-version "1.3.1")
-(def tk-version "1.4.0")
-(def tk-jetty-version "1.4.1")
-
 (defproject puppetlabs/trapperkeeper-authorization "0.7.1-SNAPSHOT"
   :description "Trapperkeeper authorization system"
   :url "http://github.com/puppetlabs/trapperkeeper-authorization"
   :license {:name "Apache License, Version 2.0"
             :url "http://www.apache.org/licenses/LICENSE-2.0.html"}
 
+  :min-lein-version "2.7.1"
+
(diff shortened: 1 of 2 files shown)


project.clj
```diff
@@ -45,16 +37,14 @@
...
+                   :dependencies [[puppetlabs/trapperkeeper-webserver-jetty9]
+                                  [puppetlabs/trapperkeeper nil :classifier "test" :scope "test"]
+                                  [puppetlabs/kitchensink nil :classifier "test" :scope "test"]]}
              :testutils {:source-paths ^:replace ["test"]}}
 
   ;; this plugin is used by jenkins jobs to interrogate the project version
-  :plugins [[lein-project-version "0.1.0"]
-            [lein-release "1.0.5" :exclusions [org.clojure/clojure]]
-            [puppetlabs/i18n "0.4.3"]]
+  :plugins [[lein-parent "0.3.1"]
```
KevinCorcoran, line 46: I think you want 0.3.3 here not 0.3.1

KevinCorcoran (changes requested):

project.clj
camlow325: I think 0.3.1 is the version I want for `lein-parent`.  That's what we're using across other projects so far.  0.3.3 is the latest for `clj-parent`, which I have in the `parent-project.coords` section above.  Wish those two versions weren't so close together.  I keep confusing the two.  😕

PR #56, opened 2017-03-29
(PE-19979) Update clj-parent to 0.6.1 and i18n to 0.8.0

haus:


camlow325 (approved):
Tests are passing so merging.

PR #58, opened 2018-11-29
WIP Pull in rbac client

Magisus:


Magisus:
Locally this is failing to find the artifact, even though other projects (e.g. pe-rbac-service) with the same dep (rbac-client 0.9.2) work fine. What am I missing?

justinstoller:
The client is closed source

Magisus:
Oh shoot, you're right, that makes sense. Hmm...

justinstoller:
I'd actually expect this to work locally on the vpn but not in travis. It looks like its failing locally because you specified a release and we only have our internal snapshot repo configured (and we are still pointing at our nexus alias which should change).

PR #59, opened 2018-12-05
(TK-475) Pull in rbac client

Magisus:
This commit pulls in the rbac client library as a dependency of the new
feature to support checking rbac permissions as part of tk-auth rules.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -6,7 +6,7 @@
 
   :min-lein-version "2.7.1"
 
-  :parent-project {:coords [puppetlabs/clj-parent "0.6.1"]
+  :parent-project {:coords [puppetlabs/clj-parent "2.4.1"]
                    :inherit [:managed-dependencies]}
 
   ;; Abort when version ranges or version conflicts are detected in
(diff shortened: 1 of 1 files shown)


project.clj
```diff
@@ -15,15 +15,20 @@
...
                  [slingshot]
                  [prismatic/schema]
-                 [ring-mock]
+                 [ring/ring-mock]
 
                  [puppetlabs/kitchensink]
                  [puppetlabs/trapperkeeper]
                  [puppetlabs/ring-middleware]
                  [puppetlabs/ssl-utils]
+                 [puppetlabs/rbac-client]
```
pcarlisle: i'd probably just bump parent here and leave this part until it's used

pcarlisle:
This breaks the repl in the example ring app, which was relying on the fact that some dependency used to pull in `org.clojure/tools.namespace` and now it no longer does. This needs to be added as an explicit dependency for the `:dev` profile.

pcarlisle (changes requested):
see comment

Magisus:
Updated. I verified that the repl can now start correctly.

PR #60, opened 2018-12-07
(TK-478) Add rbac rule to ACL schema

Magisus:
These commits update the schema for the ACL, represented as HOCON in auth.conf, to support a new map for allowing and denying.
```
allow: { rbac: { permission: "object:action:instance" } }
```

It does not currently check anything based on this rule.

### DIFF
--- a/src/puppetlabs/trapperkeeper/authorization/acl.clj
+++ b/src/puppetlabs/trapperkeeper/authorization/acl.clj
@@ -2,6 +2,7 @@
   (:require [clojure.set :refer [intersection]]
             [puppetlabs.i18n.core :refer [trs]]
             [puppetlabs.ssl-utils.core :refer [subject-alt-name-oid]]
+            [slingshot.slingshot :refer [throw+]]
             [schema.core :as schema])
   (:import clojure.lang.IFn))
 
@@ -134,9 +135,14 @@
   (cond
     ;;RBAC permission
     (map? rbac)
-    {:auth-type auth-type
-     :match :rbac-permission
-     :value (:permission rbac)}
+    (if (= :deny auth-type)
+      (throw+
+        {:kind :rbac-deny
+         :msg (trs "RBAC permissions cannot be used to deny access. Permission: ''{0}''" (:permission rbac))})
+      {:auth-type auth-type
+       :match :rbac-permission
+       :value (:permission rbac)})
+
     ;; SSL Extensions
     (map? extensions)
     {:auth-type auth-type
@@ -181,7 +187,9 @@
      :value (clojure.string/replace certname #"^/(.*)/$" "$1")}
 
     :else
-    (throw (Exception. (trs "invalid domain value: {0}" certname)))))
+    (throw+
+      {:kind :invalid-domain
+       :msg (trs "invalid domain value: {0}" certname)})))
 
 ;; ACE matching
 
--- a/src/puppetlabs/trapperkeeper/authorization/acl.clj
+++ b/src/puppetlabs/trapperkeeper/authorization/acl.clj
@@ -5,6 +5,9 @@
             [puppetlabs.i18n.core :refer [trs]]))
 
 ;; Schemas
+(def RBACRule
+  "Schema for defining an RBAC Permission"
+  {:permission schema/Str})
 
 (def OIDMap
   "Mapping of string OIDs to shortname keywords. Used to update an incoming
@@ -57,6 +60,7 @@
 (def ACEConfig
   "Schema for representing the configuration of an ACE."
   (schema/pred #(or (nil? (schema/check schema/Str (:certname %)))
+                    (nil? (schema/check RBACRule (:rbac %)))
                     (nil? (schema/check ExtensionRule (:extensions %))))
                "ACE Config Value"))
 
@@ -72,7 +76,7 @@
 (def ACE
   "An authorization entry matching a network or a domain"
   {:auth-type AuthType
-   :match (schema/enum :string :regex :backreference :extensions)
+   :match (schema/enum :string :regex :backreference :extensions :rbac-permission)
    :value ACEValue})
 
 (def ACL
@@ -125,8 +129,13 @@
 (schema/defn ^:always-validate new-domain :- ACE
   "Creates a new ACE for a domain"
   [auth-type :- AuthType
-   {:keys [certname extensions]} :- ACEConfig]
+   {:keys [certname extensions rbac]} :- ACEConfig]
   (cond
+    ;;RBAC permission
+    (map? rbac)
+    {:auth-type auth-type
+     :match :rbac-permission
+     :value (:permission rbac)}
     ;; SSL Extensions
     (map? extensions)
     {:auth-type auth-type
--- a/src/puppetlabs/trapperkeeper/authorization/acl.clj
+++ b/src/puppetlabs/trapperkeeper/authorization/acl.clj
@@ -1,13 +1,14 @@
 (ns puppetlabs.trapperkeeper.authorization.acl
   (:require [clojure.set :refer [intersection]]
-            [schema.core :as schema]
+            [puppetlabs.i18n.core :refer [trs]]
             [puppetlabs.ssl-utils.core :refer [subject-alt-name-oid]]
-            [puppetlabs.i18n.core :refer [trs]]))
+            [schema.core :as schema])
+  (:import clojure.lang.IFn))
 
 ;; Schemas
 (def RBACRule
   "Schema for defining an RBAC Permission"
-  {:permission schema/Str})
+  {:permission (schema/constrained schema/Str #(re-matches #".*:.*:.*" %))})
 
 (def OIDMap
   "Mapping of string OIDs to shortname keywords. Used to update an incoming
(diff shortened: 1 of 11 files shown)


src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -36,9 +36,13 @@
     (when-not (nil? (schema/check schema/Str (:certname config-value)))
       (reject! (trs "certname key should map to a string; got ''{0}''" (:certname config-value)))))
 
-  (when (or (not (or (:extensions config-value) (:certname config-value)))
+  (when (:rbac config-value)
+    (when-not (nil? (schema/check acl/RBACRule (:rbac config-value)))
+  (reject! (trs "permission key should map to a string; got ''{0}''" (:rbac config-value)))))
```
Magisus: Indentation is a little funny here.

src/puppetlabs/trapperkeeper/services/authorization/authorization_core.clj
```diff
@@ -36,9 +36,13 @@
...
+
+  (when (or (not (or (:extensions config-value) (:certname config-value) (:rbac config-value)))
             (and (:extensions config-value) (:certname config-value)))
-    (reject! (trs "ACL Definition must contain exactly one of ''certname'' or ''extensions'' keys; got ''{0}''" config-value))))
+    (reject! (trs "ACL Definition must contain exactly one of ''certname'' or ''extensions'' or ''permission'' keys; got ''{0}''" config-value))))
```
Magisus: I think this should say "rbac" instead of "permission", since that's the top level key and the one we're checking for.

Magisus (commented):
I think we should remove the inclusion of the service from this PR (both in the `authorization_service` file and in the test), since they're not needed for these schema updates.

Magisus:
Also looks like we might need to update Travis. I'll push something up for that.

pcarlisle:
reviewing...

test/puppetlabs/trapperkeeper/services/authorization/authorization_service_test.clj
```diff
@@ -110,7 +110,15 @@
...
     :allow-unauthenticated true
     :sort-order 500
-    :name "puppetlabs csr"}])
+    :name "puppetlabs csr"}
+   {:match-request
+    {:path "/puppet/v3/rbac_test"
+     :method "get"
+     :type "path"}
+    :allow
+      { :rbac {:permission "get:test:*" }}
```
pcarlisle: uneven formatting here
```clj
:allow {:rbac {:permission "get:test:*"}}
```
all on one line would make the most sense

src/puppetlabs/trapperkeeper/authorization/acl.clj
```diff
@@ -125,8 +129,13 @@
...
   "Creates a new ACE for a domain"
   [auth-type :- AuthType
-   {:keys [certname extensions]} :- ACEConfig]
+   {:keys [certname extensions rbac]} :- ACEConfig]
   (cond
+    (map? rbac)
+    {:auth-type auth-type
+     :match :rbac-permission
+     :value (:permission rbac)
+     }
```
pcarlisle: tuck this in

src/puppetlabs/trapperkeeper/authorization/acl.clj
```diff
@@ -125,8 +129,13 @@
 (schema/defn ^:always-validate new-domain :- ACE
   "Creates a new ACE for a domain"
   [auth-type :- AuthType
-   {:keys [certname extensions]} :- ACEConfig]
+   {:keys [certname extensions rbac]} :- ACEConfig]
   (cond
+    (map? rbac)
```
pcarlisle: every other stanza here has a comment, this one seems self explanatory but might be best for consistency

src/puppetlabs/trapperkeeper/authorization/acl.clj
```diff
@@ -57,6 +60,7 @@
 (def ACEConfig
   "Schema for representing the configuration of an ACE."
   (schema/pred #(or (nil? (schema/check schema/Str (:certname %)))
+                    (nil? (schema/check RBACRule (:rbac %)))
                     (nil? (schema/check ExtensionRule (:extensions %))))
                "ACE Config Value"))
```
pcarlisle, line 67: not the doing of this pr, but this way of using schema seems perverse

pcarlisle (approved):

pcarlisle:
The preexisting code has some weirdness, but this change is ok pending some really minor formatting bits.

src/puppetlabs/trapperkeeper/authorization/acl.clj
```diff
@@ -125,8 +129,13 @@
 (schema/defn ^:always-validate new-domain :- ACE
```
pcarlisle, line 131: It's not really a domain anymore

Magisus: Is it not? I figured "domain" here referred to "endpoint", as in, add a new rule for a "domain". Making that rule RBAC based doesn't seem to break that concept.

src/puppetlabs/trapperkeeper/authorization/acl.clj
```diff
@@ -51,12 +54,14 @@
 (def ACEChallenge
   "Pertinent authorization information extracted from a request used during
   authz rule enforcement."
-  {:certname schema/Str
-   :extensions Extensions})
+  {(schema/optional-key :certname) schema/Str
```
tvpartytonight: This change here to the `ACEChallenge` schema lets the `allowed` function use an ACE that has the rbac permission string. This didn't seem like a great solution, but it was the only I could conjure up for the time being...

src/puppetlabs/trapperkeeper/authorization/acl.clj
tvpartytonight: Domain here I think is a reference to the domain of the certname.

src/puppetlabs/trapperkeeper/authorization/acl.clj
pcarlisle: This represents the information from the request that is used to determine whether the rule works, so if anything gets added wouldn't it be the token?

test/puppetlabs/trapperkeeper/authorization/acl_test.clj
```diff
@@ -123,6 +123,11 @@
     (testing "should allow anything"
       (is (acl/allowed? acl {:certname "anything" :extensions {}})))))
 
+(deftest test-rbac-permission
+  (let [acl (acl/allow {:rbac {:permission "whee"}} )]
+                         (testing "RBAC permission string matching"
+                           (is (acl/allowed? acl { :permission "whee"} ) ))))
```
pcarlisle: Why would it be called this way?

src/puppetlabs/trapperkeeper/authorization/acl.clj
tvpartytonight: @Magisus pointed out that the extensions also doesn't necessarily work with `new-domain` either, so maybe once we decide on the appropriate name we can update it then.

test/puppetlabs/trapperkeeper/authorization/acl_test.clj
Magisus: We discussed this and decided this test is outside the scope of this change, because it involves the spot in the code where we will actually need to do the RBAC validation. This test will be reworked to test the right thing in a later PR. For now, we've removed it.

src/puppetlabs/trapperkeeper/authorization/acl.clj
Magisus: I think the comment should go one line up.

Magisus (commented):
I think this is good for the scope of this ticket.

src/puppetlabs/trapperkeeper/authorization/acl.clj
```diff
@@ -1,10 +1,14 @@
...
+            [puppetlabs.i18n.core :refer [trs]]
             [puppetlabs.ssl-utils.core :refer [subject-alt-name-oid]]
-            [puppetlabs.i18n.core :refer [trs]]))
+            [schema.core :as schema])
+  (:import clojure.lang.IFn))
 
 ;; Schemas
+(def RBACRule
+  "Schema for defining an RBAC Permission"
+  {:permission (schema/constrained schema/Str #(re-matches #".*:.*:.*" %))})
```
Magisus, line 12: Wasn't sure if we need to be more restrictive than `.` in this regex. Couldn't really find anything about the details of what characters are allowed, and I think currently the libraries don't really do any validation except that the string is splittable on `:` https://github.com/puppetlabs/clj-rbac-client/blob/master/src/puppetlabs/rbac_client/services/rbac.clj#L22.

Magisus:
This currently doesn't handle "invalid token" as a 403 (it throws and returns 500). I'll look into fixing that.

src/puppetlabs/trapperkeeper/authorization/ring_middleware.clj
```diff
@@ -234,41 +234,61 @@
...
+          (assoc request :rbac-error msg))
+        (catch [:kind :puppetlabs.rbac/invalid-token] {:keys [msg]}
+          (log/error "Failure validating RBAC token:" msg)
+          (assoc request :rbac-error msg))
+        (catch [:kind :puppetlabs.rbac/token-expired] {:keys [msg]}
+          (log/error "Failure validating RBAC token:" msg)
+          (assoc request :rbac-error msg))
+        (catch [:kind :puppetlabs.rbac-client/connection-failure] {:keys [msg]}
+          (log/error "Failure validating RBAC token:" msg)
+          (assoc request :rbac-error msg)))
```
Magisus, line 254: Do you think this is worth trying to DRY up?

Magisus:
I'm happy with the functionality of this at this point.

src/puppetlabs/trapperkeeper/authorization/ring_middleware.clj
pcarlisle: ok gave it a shot

Magisus: This works for me. TIL about the `some->` macro :)

Magisus (commented):
Can't add a review to my own PR, but 👍

PR #66, opened 2023-02-14
(maint) update dependencies, prepare for 2.0.0 release

jonathannewman:
First commit:

This updates clj-parent which brings in new versions of clj-rbac-client,
trapperkeeper, slingshot, prismatic/schema, kitchensink, ring-middleware,
ssl-urls and i18n.

Additionally ring-mock was moved into the dev-dependencies, and bouncycastle
was added as dev-dependencies for trapperkeeper-webserver-jetty9.

Clojure tools namespace was also updated to 1.4.1

At the time of the last update ring-codec was part of the ring-core
library and depended on java url decode for its implementation.  Since
then, it was moved into a separate ring-codec library and the implementation
changed to a more simple percent-encoding implementation. As a result,
there are no longer exceptions thrown by invalid encodings. A test that
was trying to detect that invalid decoding was changes to instead reflect
the new behavior.

Second commit:

This prepares for a 2.0.0 release. This release is no longer claimed
to be compatible with java 8, and thus the major version number change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: clojure
-lein: 2.8.1
+lein: 2.10.0
 jdk:
+  - openjdk17
   - openjdk11
-  - openjdk8
 script: ./ext/travisci/test.sh
 notifications:
   email: false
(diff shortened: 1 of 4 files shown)


steveax (approved):
👍

PR #67, opened 2023-08-08
(maint) fix debug logging formatting

jonathannewman:
This fixes the debug logging to correctly format the result instead of injecting a `{0}` into the logging.


steveax (approved):
👍

PR #68, opened 2023-08-08
(maint) improve warning logging message

jonathannewman:
A warning logging message was using multiple levels of formatting, and a log method that expected to format as well.  This updates the method to use one type of formatting.


steveax (approved):
👍

